PR #3, opened 2023-10-05
Add testing for the vLLM python model

pskiran1:
Changes:

- Added testing for Python backend vLLM model
- Added pre-commit hooks


dyastremsky:
Great work and tests! Based on the approach that Triton uses elsewhere, I think these tests should be in the server repo's `qa` folder with accompanying changes in GitLab to automatically run these tests. This PR replicating a lot of code in that repository, and that's the entry point for running tests for all backends at present.

Similarly, I think cloning the tutorials repo and testing them directly would be a better approach. If you copy the code over, it won't actually be testing the tutorial, since the code here will be out of step with the

dyastremsky:
> Great work and tests! Based on the approach that Triton uses elsewhere, I think these tests should be in the server repo's `qa` folder with accompanying changes in GitLab to automatically run these tests. This PR replicating a lot of code in that repository, and that's the entry point for running tests for all backends at present.
> 
> Similarly, I think cloning the tutorials repo and testing them directly would be a better approach. If you copy the code over, it won't actually be testing the tutorial, since the code here will be out of step with the tutorial code.

Discussed offline with @tanmayv25. We are working to test the backend code for now, then we can integrate this testing further in a future release.

dyastremsky (approved):

ci/qa_models/vllm/model.py
tanmayv25, line 1: This should not be needed. The model should be installed as a backend in the container. Am I missing something?

ci/qa_models/vllm/config.pbtxt
```diff
@@ -0,0 +1,68 @@
...
+# CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL,
+# EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO,
+# PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR
+# PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+name: "vllm"
+backend: "python"
```
tanmayv25, line 28: We should directly place vllm in the backend field.

pskiran1:
Closing this PR, will add testing for the tutorials separate from the backend repo.

PR #7, opened 2023-10-10
Renaming the tensors and removing tools

tanmayv25:
The changes are being cherry-picked from here: https://github.com/triton-inference-server/tutorials/pull/59


src/model.py
```diff
@@ -172,17 +174,19 @@ async def generate(self, request):
...
         try:
             request_id = random_uuid()
-            prompt = pb_utils.get_input_tensor_by_name(request, "PROMPT").as_numpy()[0]
+            prompt = pb_utils.get_input_tensor_by_name(
+                request, "text_input"
+            ).as_numpy()[0]
             if isinstance(prompt, bytes):
                 prompt = prompt.decode("utf-8")
-            stream = pb_utils.get_input_tensor_by_name(request, "STREAM").as_numpy()[0]
+            stream = pb_utils.get_input_tensor_by_name(request, "stream").as_numpy()[0]
```
nnshah1, line 182: one other change to be picked up - making stream an optional param:

https://github.com/triton-inference-server/tutorials/pull/58/files

tanmayv25: Its already included in the changes.

tanmayv25: https://github.com/triton-inference-server/vllm_backend/blob/40f38ff5c783624bf32027f8d69e77d187e028de/ci/qa_models/vllm_opt/config.pbtxt#L45

dyastremsky (approved):

dyastremsky:
Great changes. Fantastic work!

src/model.py
nnshah1: @tanmayv25 but don't we need to check here if it is included?

```suggestion
            stream = pb_utils.get_input_tensor_by_name(request, "stream")
            if stream:
               stream = stream.as_numpy()[0]
            else:
               stream = False
```

tanmayv25: Good catch! I missed it.
Probably one more test case to include in our CI tests.
@pskiran1 ^^^

PR #17, opened 2023-10-20
Extend CI tests for the vLLM backend

pskiran1:
Changes:

1. Test model loading when model.json is malformed.
2. Test Stream enabled
3. Add accuracy tests on the backend of some popular models to compare Triton's overhead.
4. Test multiple gpu utilization


ci/L0_accuracy_test/test.sh
```diff
@@ -0,0 +1,87 @@
...
+SERVER_LOG="./vllm_accuracy_test_server.log"
+CLIENT_LOG="./vllm_accuracy_test_client.log"
+TEST_RESULT_FILE='test_results.txt'
+CLIENT_PY="./vllm_accuracy_test.py"
+EXPECTED_NUM_TESTS=1
+
+rm -rf models/
+
+mkdir -p models/vllm_opt/1/
+echo '{"model":"facebook/opt-125m", "disable_log_requests": "true", "gpu_memory_utilization":0.3}' > models/vllm_opt/1/model.json
```
oandreeva-nv, line 43: Could you please adjust these tests and prior tests to re-use `samples` directory, @dyastremsky added? This way we can eliminate maintaining vllm model in qa folder

ci/L0_backend_vllm/test.sh
```diff
@@ -45,6 +47,16 @@ mkdir -p models/add_sub/1/
 cp ../qa_models/add_sub/model.py models/add_sub/1/
 cp ../qa_models/add_sub/config.pbtxt models/add_sub
 
+# Invalid model attribute
+mkdir -p models/vllm_invalid_1/1/
+cp ../qa_models/vllm_opt/config.pbtxt models/vllm_invalid_1/
```
oandreeva-nv, line 52: Same here, could you please use `samples` directory and copy file from there, then you can just change one of entries to json to an invalid attribute with `sed`: https://www.geeksforgeeks.org/sed-command-in-linux-unix-with-examples/

ci/L0_accuracy_test/vllm_accuracy_test.py
```diff
@@ -0,0 +1,122 @@
...
+            result = user_data._completed_requests.get()
+            self.assertIsNot(type(result), InferenceServerException)
+
+            output = result.as_numpy("text_output")
+            self.assertIsNotNone(output)
+
+            triton_vllm_output.extend(output)
+
+        self.triton_client.stop_stream()
+        self.assertEqual(python_vllm_output, triton_vllm_output)
```
dyastremsky, line 115: @pskiran1 Will this ever fail? I think vLLM has different outputs for the same prompt sometimes, and I don't see a seed here to make this deterministic.

ci/L0_accuracy_test/vllm_accuracy_test.py
```diff
@@ -0,0 +1,122 @@
...
+    def test_vllm_model(self):
+        user_data = UserData()
+        stream = False
+        prompts = [
+            "The most dangerous animal is",
+            "The capital of France is",
+            "The future of AI is",
+        ]
+        number_of_vllm_reqs = len(prompts)
+        sampling_parameters = {"temperature": "0.1", "top_p": "0.95"}
```
oandreeva-nv, line 84: for determinism, temperature should be set to 0 and we also need to fix random seed somewhere

ci/L0_accuracy_test/test.sh
pskiran1: Updated tests to use `samples`.

ci/L0_backend_vllm/test.sh
pskiran1: Updated tests.

ci/L0_accuracy_test/vllm_accuracy_test.py
pskiran1: Thank you. Adjusted temperature and top_p to controls the randomness of LLM output.

ci/L0_accuracy_test/vllm_accuracy_test.py
pskiran1: During the testing, performed 100 iterations and observed no failures.
However, noticed that when the temperature was set to 0.5, it resulted in a failure.
To eliminate randomness and achieve consistent LLM output, adjusted the temperature to 0 and top_p to 1.
Conducted the test for 100 iterations again to ensure that there was no failure.

dyastremsky: Great work!

dyastremsky (dismissed):
Great work on this testing!

ci/L0_accuracy_test/test.sh
```diff
@@ -0,0 +1,86 @@
...
+CLIENT_PY="./vllm_accuracy_test.py"
+SAMPLE_MODELS_REPO="../../samples/model_repository"
+EXPECTED_NUM_TESTS=1
+
+rm -rf models && mkdir -p models
+cp -r ${SAMPLE_MODELS_REPO}/vllm_model models/vllm_opt
+sed -i 's/"gpu_memory_utilization": 0.5/"gpu_memory_utilization": 0.3/' models/vllm_opt/1/model.json
+
+pip3 install tritonclient
+pip3 install grpcio
```
oandreeva-nv, line 46: nit: we can install everything in one line

ci/L0_multi_gpu/test.sh
```diff
@@ -0,0 +1,87 @@
...
+SAMPLE_MODELS_REPO="../../samples/model_repository"
+EXPECTED_NUM_TESTS=1
+
+rm -rf models && mkdir -p models
+cp -r ${SAMPLE_MODELS_REPO}/vllm_model models/vllm_opt
+sed -i '3s/^/    "tensor_parallel_size": 2,\n/' models/vllm_opt/1/model.json
+
+pip3 install tritonclient
+pip3 install grpcio
+pip3 install nvidia-ml-py3
```
oandreeva-nv, line 47: ditto

ci/L0_stream_enabled/test.sh
```diff
@@ -0,0 +1,85 @@
...
+TEST_RESULT_FILE='test_results.txt'
+CLIENT_PY="./vllm_stream_enabled_test.py"
+SAMPLE_MODELS_REPO="../../samples/model_repository"
+EXPECTED_NUM_TESTS=1
+
+rm -rf models && mkdir -p models
+cp -r ${SAMPLE_MODELS_REPO}/vllm_model models/vllm_opt
+
+pip3 install tritonclient
+pip3 install grpcio
```
oandreeva-nv, line 45: ditto

oandreeva-nv:
Nice work on tests. I have one point to discuss: I don't think we need this fine granularity in tests. I think that everything, what is using 1 gpu, can go into L0_vllm_tests and probably sub-folders, similar to how we have a set up for python backend: https://github.com/triton-inference-server/server/tree/main/qa/L0_backend_python

Then, we can keep muti-GPU tests separately, so that the time of reserving 4 GPU runner in dlcluster does not impact 1-GPU tests

oandreeva-nv (approved):

PR #19, opened 2023-10-30
Request cancellation test

pskiran1:


ci/L0_backend_vllm/request_cancel/request_cancel_test.py
```diff
@@ -0,0 +1,78 @@
+# Copyright 2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
dyastremsky, line 1: I don't love this naming of sub tests here. Could you look at what is used in server/qa and see if there is a way to be consistent with that naming? We'd also want to be internally consistent. For example, "request_cancellation" and "enabled_stream" so that they are all nouns.

ci/L0_backend_vllm/request_cancel/test.sh
```diff
@@ -0,0 +1,85 @@
...
+SERVER_LOG="./request_cancel_server.log"
+CLIENT_LOG="./request_cancel_client.log"
+TEST_RESULT_FILE='test_results.txt'
+CLIENT_PY="./request_cancel_test.py"
+SAMPLE_MODELS_REPO="../../../samples/model_repository"
+EXPECTED_NUM_TESTS=1
+
+rm -rf models && mkdir -p models
+cp -r ${SAMPLE_MODELS_REPO}/vllm_model models/vllm_opt
+sed -i 's/"model":"facebook\/opt-125m"/"model": "facebook\/opt-350m"/' models/vllm_opt/1/model.json
```
dyastremsky, line 43: Why is this necessary?

ci/L0_multi_gpu/test.sh
```diff
@@ -42,7 +42,7 @@ rm -rf models && mkdir -p models
 cp -r ${SAMPLE_MODELS_REPO}/vllm_model models/vllm_opt
 sed -i '3s/^/    "tensor_parallel_size": 2,\n/' models/vllm_opt/1/model.json
 
-pip3 install tritonclient grpcio nvidia-ml-py3
+python3 -m pip install --upgrade pip && pip3 install tritonclient[all] nvidia-ml-py3
```
dyastremsky, line 45: Do you need both clients or just the grpc client in `tritonclient[all]`?

dyastremsky (commented):
Great work! Commented a few questions.

ci/L0_multi_gpu/test.sh
pskiran1: Thank you for pointing this out. We need only GPC. Modified it to `tritonclient[grpc]`.

ci/L0_backend_vllm/request_cancel/request_cancel_test.py
pskiran1: Thank you, updated subtests names.

ci/L0_backend_vllm/request_cancel/test.sh
pskiran1: Thank you, this change is not necessary. Missed to undo some intermediate change.

dyastremsky (approved):

ci/L0_multi_gpu/test.sh
```diff
@@ -42,7 +42,7 @@ rm -rf models && mkdir -p models
 cp -r ${SAMPLE_MODELS_REPO}/vllm_model models/vllm_opt
 sed -i '3s/^/    "tensor_parallel_size": 2,\n/' models/vllm_opt/1/model.json
 
-pip3 install tritonclient grpcio nvidia-ml-py3
+python3 -m pip install --upgrade pip && pip3 install tritonclient[grpc] nvidia-ml-py3
```
dyastremsky, line 45: One more quick comment: for this and the other pip upgrade... is this necessary? It looks like the QA container should have [the latest pip](https://github.com/triton-inference-server/server/blob/d33eaa5a153f5333374046f12826aac5ffd5ed52/Dockerfile.QA#L62) already. Is this something that needs to be added to the vllm backend container instead?

ci/L0_backend_vllm/request_cancellation/request_cancellation_test.py
```diff
@@ -0,0 +1,78 @@
...
+            user_data = UserData()
+            model_name = "vllm_opt"
+            stream = False
+            sampling_parameters = {"temperature": "0.75", "top_p": "0.9"}
+
+            triton_client.start_stream(callback=partial(callback, user_data))
+
+            for i in range(100):
+                prompt = (
+                    f"Write an original and creative poem of at least {100 + i} words."
```
tanmayv25, line 50: Is there a way to test whether the request cancellation was aborted by the backend when running the requests iteratively in the [loop](https://github.com/triton-inference-server/vllm_backend/blob/main/src/model.py#L212)?
May be add a verbose log when calling abort and search for its presence.
I believe this prompt is an attempt to keep the requests under processing when cancellation is invoked. I would assume sending a single long-running request and then invoking the cancellation, proceeded by checking for an abort should be sufficient to detect the cancellation.

ci/L0_multi_gpu/test.sh
tanmayv25: @dyastremsky  I would advise not to include the additional libraries in our vLLM backend container that we release on NGC as it is expected to be a runtime image similar to our full-fat container image.
For simplicity, we didn't create a separate QA container as we do in our general container CI. 
We might have done that if there were significant dependencies and setup that were needed to be conducted.

tanmayv25: I would suggest let's start from here and see how it goes.. We may consider using a common requirements.txt file to install packages in several tests.. Or go with a separate QA container build in our pipeline.

dyastremsky: Okay, sounds good! This just seems like the wrong place to have the responsibility for upgrading pip. We can do it now for a less-scalable option and revisit creating a vLLM-QA container, if needed later.

ci/L0_backend_vllm/request_cancellation/request_cancellation_test.py
pskiran1: Thank you, updated the test with the suggested changes.

tanmayv25 (approved):

PR #20, opened 2023-11-03
Adding autocomplete to vllm model.py

oandreeva-nv:
Moving default parameters from config to auto_complete function.

Also tests Python_backend's `set_model_transaction_policy`

### DIFF
--- a/src/model.py
+++ b/src/model.py
@@ -41,49 +41,6 @@ _VLLM_ENGINE_ARGS_FILENAME = "model.json"
 
 
 class TritonPythonModel:
-    @staticmethod
-    def auto_complete_config(auto_complete_model_config):
-        inputs = [
-            {"name": "text_input", "data_type": "TYPE_STRING", "dims": [1]},
-            {"name": "stream", "data_type": "TYPE_BOOL", "dims": [1]},
-            {
-                "name": "sampling_parameters",
-                "data_type": "TYPE_STRING",
-                "dims": [1],
-                "optional": True,
-            },
-        ]
-        outputs = [{"name": "text_output", "data_type": "TYPE_STRING", "dims": [-1]}]
-
-        # Store the model configuration as a dictionary.
-        config = auto_complete_model_config.as_dict()
-        input_names = []
-        output_names = []
-        for input in config["input"]:
-            input_names.append(input["name"])
-        for output in config["output"]:
-            output_names.append(output["name"])
-
-        # Add only missing inputs and output to the model configuration.
-        for input in inputs:
-            if input["name"] not in input_names:
-                auto_complete_model_config.add_input(input)
-        for output in outputs:
-            if output["name"] not in output_names:
-                auto_complete_model_config.add_output(output)
-
-        # We need to use decoupled transaction policy for saturating
-        # vLLM engine for max throughtput.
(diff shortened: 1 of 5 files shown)


src/model.py
```diff
@@ -41,6 +41,49 @@
 
 
 class TritonPythonModel:
+    @staticmethod
+    def auto_complete_config(auto_complete_model_config):
+        inputs = [
+            {"name": "text_input", "data_type": "TYPE_STRING", "dims": [1]},
+            {"name": "stream", "data_type": "TYPE_BOOL", "dims": [1]},
```
rmccorm4, line 48: stream should be optional too IMO. I think it was in the tutorial.

src/model.py
```diff
@@ -41,6 +41,49 @@
...
+        for output in config["output"]:
+            output_names.append(output["name"])
+
+        # Add only missing inputs and output to the model configuration.
+        for input in inputs:
+            if input["name"] not in input_names:
+                auto_complete_model_config.add_input(input)
+        for output in outputs:
+            if output["name"] not in output_names:
+                auto_complete_model_config.add_output(output)
```
rmccorm4, line 73: optional: I think this could be condensed a bit if you wanted to
```python
config = auto_complete_model_config.as_dict()
config_inputs = [i["name"] for i in config["input"]]
config_outputs = [o["name"] for o in config["output"]]
[auto_complete_model_config.add_input(i) for i in inputs if i["name"] not in config_inputs]
[auto_complete_model_config.add_output(o) for o in outputs if o["name"] not in config_outputs]
```

src/model.py
oandreeva-nv: I'll double check and update

src/model.py
oandreeva-nv: I would like to keep it as clear as possible, since this is a reference point from the tutorial (I mean the whole `model.py`), unless this impact performance

src/model.py
oandreeva-nv: Nice catch, for some reason during the migration from the tutorial to here, `optional` in "stream" got lost, I'll update accordingly.

src/model.py
rmccorm4: Sounds good to me

src/model.py
oandreeva-nv: added optional strem

dyastremsky (dismissed):

rmccorm4 (dismissed):

krishung5 (dismissed):

tanmayv25 (approved):

tanmayv25 (approved):

rmccorm4 (approved):

rmccorm4:
@tanmayv25 should we merge this?

tanmayv25:
I was unable to get a successful run using this change with olga's pipeline. I am in middle of investigating what is not working.

tanmayv25:
Seems like there was some setup problem. Tested successfully with job id 73883232.

rguo123:
Does autocomplete work? It doesn't seem to be called anywhere in the code besides the method being defined?

tanmayv25:
The auto-complete feature is currently available only on triton's main branch that tracks development to a future release. The auto-complete functionality will be available with 23.11 release which is scheduled to be released by end of the month.
Thanks for your patience.

PR #31, opened 2024-02-05
Why prepending output with prompt?

mkhludnev:
I'd rather created issue for discussion, but this repo doesn't have issues enabled.  First of all such prepending seems redundant especially for long RAG prompts. Then, it's an actual problem since I notice that triton gRPC crops the long response. Curiously, REST doesn't crop payload and full concatenation of prompt and output arrives to client. 

Here I put more details of the issue https://github.com/langchain-ai/langchain/issues/12474#issuecomment-1925914483

### DIFF
--- a/src/model.py
+++ b/src/model.py
@@ -213,7 +213,7 @@ class TritonPythonModel:
         """
         prompt = vllm_output.prompt
         text_outputs = [
-            (prompt + output.text).encode("utf-8") for output in vllm_output.outputs
+            output.text.encode("utf-8") for output in vllm_output.outputs
         ]
         triton_output_tensor = pb_utils.Tensor(
             "text_output", np.asarray(text_outputs, dtype=self.output_dtype)


mkhludnev:
@oandreeva-nv hello, may I ask you a question? 👆

oandreeva-nv:
Hi @mkhludnev, our ReadMe mentions that you can report all issues on [issues page](https://github.com/triton-inference-server/server/issues), which is for server. Apologies for confusion, we use server repo as the main entry-point for questions & issues.

Regarding your question, I believe the main goal is to report the full text prompt + generated portion. Do you happen to have a reproducer for gRPC cropping the part of the output? If it does, I am concerned that simply removing `prompt` from the response may not work

mkhludnev:
Considering https://github.com/triton-inference-server/server/discussions/6867 it's worth to implement `echo=False` option for vlm and trt backends.

PR #35, opened 2024-02-27
Add `exclude_input_in_output` option to vllm backend

oandreeva-nv:
This PR adds `exclude_input_in_output` flag to vllm backend inputs.
It only impacts non-streaming case.
For streaming case, I refactored code and return only `diffs`, e.g.:
Prompt = "The most dangerous animal is"
Response:
```
" the",
            " one",
            " that",
            " is",
            " most",
            " likely",
            " to",
            " be",
            " killed",
            " by",
            " a",
            " car",
            ".",
```
The above case will be the only possible response in the streaming mode. Open to discussions.
cc @@mkhludnev

### DIFF
--- a/src/model.py
+++ b/src/model.py
@@ -61,7 +61,7 @@ class TritonPythonModel:
                 "name": "exclude_input_in_output",
                 "data_type": "TYPE_BOOL",
                 "dims": [1],
-                "optional": False,
+                "optional": True,
             },
         ]
         outputs = [{"name": "text_output", "data_type": "TYPE_STRING", "dims": [-1]}]
@@ -212,13 +212,13 @@ class TritonPythonModel:
 
         return params_dict
 
-    def create_response(self, vllm_output, exclude_input_in_output):
+    def create_response(self, vllm_output, prepend_input):
         """
         Parses the output from the vLLM engine into Triton
         response.
         """
         prompt = ""
-        if not exclude_input_in_output:
+        if prepend_input:
             prompt = vllm_output.prompt
         text_outputs = [
             (prompt + output.text).encode("utf-8") for output in vllm_output.outputs
@@ -228,17 +228,19 @@ class TritonPythonModel:
         )
         return pb_utils.InferenceResponse(output_tensors=[triton_output_tensor])
 
-    def create_stream_response(self, vllm_output, last_outputs):
+    def create_stream_response(self, vllm_output, previous_outputs_lengths):
         """
         Parses the output from the vLLM engine, extracts only newly generated
         text and packs it into Triton response.
         """
-        if last_outputs is None:
-            return self.create_response(vllm_output, exclude_input_in_output=True)
+        if previous_outputs_lengths is None:
+            return self.create_response(vllm_output, prepend_input=False)
 
         text_outputs = [
-            (output.text[len(last_output.text) :]).encode("utf-8")
-            for output, last_output in zip(vllm_output.outputs, last_outputs.outputs)
+            (output.text[prev_output_length:]).encode("utf-8")
+            for output, prev_output_length in zip(
+                vllm_output.outputs, previous_outputs_lengths
+            )
         ]
         triton_output_tensor = pb_utils.Tensor(
             "text_output", np.asarray(text_outputs, dtype=self.output_dtype)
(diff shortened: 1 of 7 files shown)


mkhludnev:
Makes sense. Thank you so much @oandreeva-nv! Looking forward for release.

oandreeva-nv:
@mkhludnev This most likely will be a part of 24.03 release, but you don't need to wait this long.
For this PR, you can simply replace `/opt/tritonserver/backends/vllm/model.py` with the updated version in 24.01 or 24.02 (soon to be released), and it should work, since there's no related changes in `tritonserver` itself, to handle this flag.

src/model.py
```diff
@@ -238,6 +263,21 @@ async def generate(self, request):
...
+                request, "exclude_input_in_output"
+            )
+            if exclude_input_in_output:
+                exclude_input_in_output = exclude_input_in_output.as_numpy()[0]
+            elif exclude_input_in_output is None and stream:
+                exclude_input_in_output = True
+            else:
+                exclude_input_in_output = False
+
+            self.logger.log_info(
```
mkhludnev, line 276: I dont' see any other params logged per request. Wouldn't it be too verbose?

oandreeva-nv: Good point, How about we leave this entry for the case when `stream` is `True`, but `exclude_input_in_output` is intentionally set to `False`. For `stream` case, we never include prompt in this logic, but it would be nice to notify users that `you tried to include prompt, but we will not`

oandreeva-nv: updated

src/model.py
```diff
@@ -57,6 +57,12 @@ def auto_complete_config(auto_complete_model_config):
                 "dims": [1],
                 "optional": True,
             },
+            {
+                "name": "exclude_input_in_output",
+                "data_type": "TYPE_BOOL",
+                "dims": [1],
+                "optional": False,
```
nnshah1, line 64: why not optional?

oandreeva-nv: Good catch

src/model.py
```diff
@@ -220,6 +228,23 @@ def create_response(self, vllm_output):
         )
         return pb_utils.InferenceResponse(output_tensors=[triton_output_tensor])
 
+    def create_stream_response(self, vllm_output, last_outputs):
+        """
+        Parses the output from the vLLM engine, extracts only newly generated
+        text and packs it into Triton response.
+        """
+        if last_outputs is None:
```
nnshah1, line 236:
```suggestion
        if previous_outputs is None:
```
Small suggestion here, in another context I was reminded that "last" can also be confused with "final", "previous" makes a little clearer - but just a suggestion.

src/model.py
```diff
@@ -220,6 +228,23 @@ def create_response(self, vllm_output):
...
+            return self.create_response(vllm_output, exclude_input_in_output=True)
+
+        text_outputs = [
+            (output.text[len(last_output.text) :]).encode("utf-8")
+            for output, last_output in zip(vllm_output.outputs, last_outputs.outputs)
```
nnshah1, line 241: it occurs to me - you could store the current length of the previous output instead of the output itself - if its not needed elsewhere

src/model.py
```diff
@@ -238,6 +263,22 @@ async def generate(self, request):
                 stream = stream.as_numpy()[0]
             else:
                 stream = False
+            exclude_input_in_output = pb_utils.get_input_tensor_by_name(
+                request, "exclude_input_in_output"
+            )
+            if exclude_input_in_output:
+                exclude_input_in_output = exclude_input_in_output.as_numpy()[0]
+            elif exclude_input_in_output is None and stream:
+                exclude_input_in_output = True
```
nnshah1, line 272: just so I understand, default in stream case is True, default in non stream case is False ?

src/model.py
```diff
@@ -238,6 +263,22 @@ async def generate(self, request):
...
+            else:
+                exclude_input_in_output = False
+
+            if not exclude_input_in_output and stream:
+                exclude_input_in_output = True
```
nnshah1, line 277: I think I would error here instead of switching it -

src/model.py
oandreeva-nv: yes, that is correct

src/model.py
```diff
@@ -238,6 +263,22 @@ async def generate(self, request):
                 stream = stream.as_numpy()[0]
             else:
                 stream = False
+            exclude_input_in_output = pb_utils.get_input_tensor_by_name(
```
nnshah1, line 266:
```suggestion
            prepend_input = not pb_utils.get_input_tensor_by_name(
```

I think in this case - I would recommend renaming the variable internally with a note.

src/model.py
rmccorm4: To clarify my understanding - we'll pick the correct default if this flag is left unspecified (optional) - but we'll error check it if specified?

oandreeva-nv: > we'll pick the correct default if this flag is left unspecified (optional)

yes

> but we'll error check it if specified

at this point the error is only possible, if `exclude_input_in_output` was provided by user and `stream` was provided by user.

nnshah1 (approved):
Nice work!

mkhludnev:
A liitle bit of context https://github.com/triton-inference-server/server/discussions/6867 
Thanks, @oandreeva-nv so much!!

PR #42, opened 2024-05-24
fix: Enhance checks around KIND_GPU and tensor parallelism

rmccorm4:
##### Problem

When loading multiple instances of a vLLM model on a multi-gpu system (default behavior with KIND_GPU which is the default instance group when left unspecified), all model instances will default to the same device and can cause a CUDA OOM rather than loading a model on each GPU device assigned by the instance group settings.

This is rooted in Triton's KIND_GPU behavior making an assumption that the model is assigned only one GPU. In the future, KIND_GPU may be expanded to define a set of multiple GPUs for a single model. However, for now the recommendation is to use KIND_MODEL when a model can have multiple GPUs and use them freely (such as python models).

##### Solution

These changes try to account for this assumption by isolating the assigned GPU device ID when KIND_GPU is used, and otherwise raising an error and recommending usage of KIND_MODEL when the vLLM config implies that this is a multi-GPU model (such as `tensor_parallel_size > 1`).

#### Checklist

#### Commit Type:
Check the [conventional commit type](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type) box here and add the label to the github PR.
- [ ] build
- [ ] chore
- [ ] ci
- [ ] docs
- [ ] feat
- [x] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [x] test

#### Test plan

1. Single-gpu model on a multi-gpu system with KIND_GPU (default)

**Change**: Success, specify device ID assigned by Triton Core (from the instance group) when initializing vLLM to avoid OOM from all instances defaulting to device 0.

```
I0524 19:05:41.036859 3948 model.py:170] Detected KIND_GPU model instance, explicitly setting GPU device=0 for llama-3-8b-instruct_0
I0524 19:05:41.043640 3948 model.py:170] Detected KIND_GPU model instance, explicitly setting GPU device=1 for llama-3-8b-instruct_1
```

2. Multi-gpu model (tensor_parallel_size > 1) with KIND_GPU (default)

**Change**: Failure, with clear error to specify KIND_MODEL instead for multi-gpu models

image

3. Multi-gpu model (tensor_parallel_size > 1) with KIND_MODEL (manually specified)

Success, and auto uses Ray Worker.

image

#### Caveats

1. If there are other vLLM config fields that can imply multi-gpu other than `tensor_parallel_size`, we can add checks for those too.
2. For `KIND_MODEL` models with `tensor_parallelism == 1` and `model_instance_count > 1`, we run into the same issue where vLLM will try to allocate each instance on the same GPU. This could be enhanced with a similar check in this PR, or  deferred to future enhancement.
3. This PR doesn't provide an explicitly good way to support the case for multiple multi-gpu instances.
    - Take a 2-GPU model that you want 4 copies of on an 8-GPU system as an example.
    - If you specify KIND_MODEL in the config with 4 model instances, and `tensor_parallel_size: 2`, with 8-GPUs at your disposal, there's currently no explicit check or validation around this. I don't know how this would behave, and would likely default to how the RayWorker logic in vLLM would attempt to assign GPUs.

### DIFF
--- a/ci/L0_multi_gpu/vllm_backend/test.sh
+++ b/ci/L0_multi_gpu/vllm_backend/test.sh
@@ -31,52 +31,105 @@ TRITON_DIR=${TRITON_DIR:="/opt/tritonserver"}
 SERVER=${TRITON_DIR}/bin/tritonserver
 BACKEND_DIR=${TRITON_DIR}/backends
 SERVER_ARGS="--model-repository=`pwd`/models --backend-directory=${BACKEND_DIR} --model-control-mode=explicit --log-verbose=1"
-SERVER_LOG="./vllm_multi_gpu_test_server.log"
-CLIENT_LOG="./vllm_multi_gpu_test_client.log"
 TEST_RESULT_FILE='test_results.txt'
 CLIENT_PY="./vllm_multi_gpu_test.py"
 SAMPLE_MODELS_REPO="../../../samples/model_repository"
 EXPECTED_NUM_TESTS=1
 
-rm -rf models && mkdir -p models
-cp -r ${SAMPLE_MODELS_REPO}/vllm_model models/vllm_opt
-sed -i '3s/^/    "tensor_parallel_size": 2,\n/' models/vllm_opt/1/model.json
+### Helpers
+function validate_file_contains() {
+    local KEY="${1}"
+    local FILE="${2}"
 
-python3 -m pip install --upgrade pip && pip3 install tritonclient[grpc] nvidia-ml-py3
+    if [ -z "${KEY}" ] || [ -z "${FILE}" ]; then
+        echo "Error: KEY and FILE must be provided."
+        return 1
+    fi
 
-RET=0
+    if [ ! -f "${FILE}" ]; then
+        echo "Error: File '${FILE}' does not exist."
+        return 1
+    fi
 
-run_server
-if [ "$SERVER_PID" == "0" ]; then
-    cat $SERVER_LOG
-    echo -e "\n***\n*** Failed to start $SERVER\n***"
-    exit 1
-fi
+    count=$(grep -o -w "${KEY}" "${FILE}" | wc -l)
 
-set +e
-python3 $CLIENT_PY -v > $CLIENT_LOG 2>&1
+    if [ "${count}" -ne 1 ]; then
+        echo "Error: KEY '${KEY}' found ${count} times in '${FILE}'. Expected exactly once."
+        return 1
+    fi
+}
 
-if [ $? -ne 0 ]; then
-    cat $CLIENT_LOG
-    echo -e "\n***\n*** Running $CLIENT_PY FAILED. \n***"
-    RET=1
-else
-    check_test_results $TEST_RESULT_FILE $EXPECTED_NUM_TESTS
+function run_multi_gpu_test() {
+    export KIND="${1}"
+    export TENSOR_PARALLELISM="${2}"
+    export INSTANCE_COUNT="${3}"
+
+    # Setup a clean model repository
+    export TEST_MODEL="vllm_opt_${KIND}_tp${TENSOR_PARALLELISM}_count${INSTANCE_COUNT}"
+    local TEST_MODEL_TRITON_CONFIG="models/${TEST_MODEL}/config.pbtxt"
+    local TEST_MODEL_VLLM_CONFIG="models/${TEST_MODEL}/1/model.json"
+
+    rm -rf models && mkdir -p models
+    cp -r "${SAMPLE_MODELS_REPO}/vllm_model" "models/${TEST_MODEL}"
+    sed -i "s/KIND_MODEL/${KIND}/" "${TEST_MODEL_TRITON_CONFIG}"
+    sed -i "3s/^/    \"tensor_parallel_size\": ${TENSOR_PARALLELISM},\n/" "${TEST_MODEL_VLLM_CONFIG}"
+    # Assert the correct kind is set in case the template config changes in the future
+    validate_file_contains "${KIND}" "${TEST_MODEL_TRITON_CONFIG}"
+
+    # Start server
+    echo "Running multi-GPU test with kind=${KIND}, tp=${TENSOR_PARALLELISM}, instance_count=${INSTANCE_COUNT}"
+    SERVER_LOG="./vllm_multi_gpu_test--${KIND}_tp${TENSOR_PARALLELISM}_count${INSTANCE_COUNT}--server.log"
+    run_server
+    if [ "$SERVER_PID" == "0" ]; then
+        cat $SERVER_LOG
+        echo -e "\n***\n*** Failed to start $SERVER\n***"
+        exit 1
+    fi
+
+    # Run unit tests
+    set +e
+    CLIENT_LOG="./vllm_multi_gpu_test--${KIND}_tp${TENSOR_PARALLELISM}_count${INSTANCE_COUNT}--client.log"
+    python3 $CLIENT_PY -v > $CLIENT_LOG 2>&1
+
     if [ $? -ne 0 ]; then
         cat $CLIENT_LOG
-        echo -e "\n***\n*** Test Result Verification FAILED.\n***"
+        echo -e "\n***\n*** Running $CLIENT_PY FAILED. \n***"
         RET=1
+    else
+        check_test_results $TEST_RESULT_FILE $EXPECTED_NUM_TESTS
+        if [ $? -ne 0 ]; then
+            cat $CLIENT_LOG
+            echo -e "\n***\n*** Test Result Verification FAILED.\n***"
+            RET=1
+        fi
     fi
-fi
-set -e
+    set -e
 
-kill $SERVER_PID
-wait $SERVER_PID
-rm -rf models/
+    # Cleanup
+    kill $SERVER_PID
+    wait $SERVER_PID
+}
 
+### Test
+python3 -m pip install --upgrade pip && pip3 install tritonclient[grpc] nvidia-ml-py3
+rm -f *.log
+RET=0
+
+# Test the various cases of kind, tensor parallelism, and instance count
+# for different ways to run multi-GPU models with vLLM on Triton
+KINDS="KIND_MODEL KIND_GPU"
+TPS="1 2"
+INSTANCE_COUNTS="1 2"
+for kind in ${KINDS}; do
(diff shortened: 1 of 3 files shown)


rmccorm4:
Didn't mean for the Lora stuff to be removed - will fix that.

oandreeva-nv:
I need LoRA back before I can approve

GuanLuo (dismissed):

GuanLuo (approved):

src/model.py
```diff
@@ -146,21 +173,31 @@ def initialize(self, args):
...
-        self._loop_thread.start()
+        # If KIND_GPU is specified, isolate the selected GPU device to ensure that multiple model instances do not all use the same default
+        # device (usually device 0) when KIND_GPU is used.
+        if triton_kind == "GPU" and int(triton_device_id) >= 0:
+            self.logger.log_info(
+                f"Detected KIND_GPU model instance, explicitly setting GPU device={triton_device_id} for {triton_instance}"
+            )
+            # NOTE: this only affects this process and it's subprocesses, not other processes.
+            # vLLM doesn't currently seem to expose selecting a specific device in the APIs.
+            os.environ["CUDA_VISIBLE_DEVICES"] = triton_device_id
```
oandreeva-nv, line 200: I'm going to leave my observations here for the record.

I've tried also [`LOCAL_RANK` ](https://github.com/vllm-project/vllm/blob/e64fde4b013cb8bb2321f59ba78aca50b02071cb/vllm/envs.py#L138). It is very flaky and  GPU block's calculations do not correspond to what is calculated in "GPU-isolated" case.

oandreeva-nv (approved):
lgtm, my only ask is to try re-running tests a couple of times, to see if any flakiness is happenning.

src/model.py
```diff
@@ -146,21 +173,31 @@ def initialize(self, args):
...
-        self.output_dtype = pb_utils.triton_string_to_numpy(output_config["data_type"])
-
-        # Counter to keep track of ongoing request counts
-        self.ongoing_request_count = 0
+    def validate_device_config(self):
+        triton_kind = self.args["model_instance_kind"]
+        triton_device_id = self.args["model_instance_device_id"]
+        triton_instance = f"{self.args['model_name']}_{triton_device_id}"
+
+        # Triton's current definition of KIND_GPU makes assumptions that
```
nnshah1, line 181: @rmccorm4 just want to confirm this is documented in the description of KIND_GPU in the docs

src/model.py
```diff
@@ -146,21 +173,31 @@ def initialize(self, args):
...
-        self._shutdown_event = asyncio.Event()
-        self._loop_thread.start()
+        # If KIND_GPU is specified, isolate the selected GPU device to ensure that multiple model instances do not all use the same default
+        # device (usually device 0) when KIND_GPU is used.
+        if triton_kind == "GPU" and int(triton_device_id) >= 0:
+            self.logger.log_info(
+                f"Detected KIND_GPU model instance, explicitly setting GPU device={triton_device_id} for {triton_instance}"
+            )
+            # NOTE: this only affects this process and it's subprocesses, not other processes.
+            # vLLM doesn't currently seem to expose selecting a specific device in the APIs.
```
nnshah1, line 199:
```suggestion
            # vLLM currently recommends setting CUDA VISIBLE DEVICES to select GPUs
```

nnshah1: From what we can see in their code, examples, and comments - this is the current recommended way to select GPUs for vllm

nnshah1 (approved):
Thanks for getting the fix in quickly!

Tabrizian:
I think we want to allow the model to interact with other GPUs even if the KIND_GPU and a specific device is specified. The reason is that the model can be part of an ensemble pipeline and it may want to copy the tensors from other devices even though the actual execution is happening on device_id. 

I think it is better to set the default context to the device id specified by Triton. For vllm, this can be achieved using `torch.cuda.set_device`.

src/model.py
rmccorm4: https://docs.nvidia.com/deeplearning/triton-inference-server/user-guide/docs/user_guide/model_configuration.html#multiple-model-instances

> By default, a single execution instance of the model is created for each GPU available in the system.

src/model.py
```diff
@@ -146,21 +174,31 @@ def initialize(self, args):
...
-        )
-        self._shutdown_event = asyncio.Event()
-        self._loop_thread.start()
+        # If KIND_GPU is specified, specify the device ID assigned by Triton to ensure that
+        # multiple model instances do not oversubscribe the same default device.
+        if triton_kind == "GPU" and triton_device_id >= 0:
+            self.logger.log_info(
+                f"Detected KIND_GPU model instance, explicitly setting GPU device={triton_device_id} for {triton_instance}"
+            )
+            # vLLM doesn't currently expose device selection in the APIs
```
oandreeva-nv, line 200:
```suggestion
            # vLLM doesn't currently (v0.4.2) expose device selection in the APIs
```

ci/L0_multi_gpu/vllm_backend/test.sh
```diff
@@ -31,50 +31,104 @@ TRITON_DIR=${TRITON_DIR:="/opt/tritonserver"}
...
+    if [ "${count}" -ne 1 ]; then
+        echo "Error: KEY '${KEY}' found ${count} times in '${FILE}'. Expected exactly once."
+        return 1
+    fi
+}
+
+function run_multi_gpu_test() {
+    export KIND="${1}"
+    export TENSOR_PARALLELISM="${2}"
+    export INSTANCE_COUNT="${3}"
```
GuanLuo, line 65: Do you kneed `export`? Looks like all usages are local

ci/L0_multi_gpu/vllm_backend/test.sh
```diff
@@ -31,50 +31,104 @@ TRITON_DIR=${TRITON_DIR:="/opt/tritonserver"}
...
+    if [ "$SERVER_PID" == "0" ]; then
+        cat $SERVER_LOG
+        echo -e "\n***\n*** Failed to start $SERVER\n***"
+        exit 1
+    fi
+
+    # Run unit tests
+    set +e
+    CLIENT_LOG="./vllm_multi_gpu_test--${KIND}_tp${TENSOR_PARALLELISM}_count${INSTANCE_COUNT}--client.log"
+    python3 $CLIENT_PY -v > $CLIENT_LOG 2>&1
```
GuanLuo, line 92: Running all unit tests against different settings? Is that necessary?

ci/L0_multi_gpu/vllm_backend/test.sh
GuanLuo: Okay I see it now.. Should try to move server setup into py test (setup / teardown), @jbkyang-nvi had done something similar.

ci/L0_multi_gpu/vllm_backend/vllm_multi_gpu_test.py
```diff
@@ -118,6 +123,59 @@ def _test_vllm_model(self, send_parameters_as_tensor=True):
...
+        print(f"Test Matrix: {model=}, {kind=}, {tp=}, {instance_count=}")
+
+        # Only support tensor parallelism or multiple instances for now, but not both.
+        # Support for multi-instance tensor parallel models may require more
+        # special handling in the backend to better handle device assignment.
+        # NOTE: This eliminates the 1*1=1 and 2*2=4 test cases.
+        if int(tp) * int(instance_count) != 2:
+            msg = "TENSOR_PARALLELISM and INSTANCE_COUNT must have a product of 2 for this 2-GPU test"
+            print("Skipping Test:", msg)
+            self.skipTest(msg)
```
oandreeva-nv, line 156: Can we put this into `@unittest.skipIf` ? It would be easier to locate then

ci/L0_multi_gpu/vllm_backend/vllm_multi_gpu_test.py
```diff
@@ -118,6 +123,59 @@ def _test_vllm_model(self, send_parameters_as_tensor=True):
...
+            return
+
+        # Loading a KIND_MODEL model with multiple instances can cause
+        # oversubscription to specific GPUs and cause a CUDA OOM if the
+        # gpu_memory_utilization settings are high without further handling
+        # of device assignment in the backend.
+        if kind == "KIND_MODEL" and int(instance_count) > 1:
+            msg = "Testing multiple model instances of KIND_MODEL is not implemented at this time"
+            print("Skipping Test:", msg)
+            self.skipTest(msg)
```
oandreeva-nv, line 175: ditto

ci/L0_multi_gpu/vllm_backend/vllm_multi_gpu_test.py
rmccorm4: I think I'd have to move the tp and instance_counts to be global or passed directly to the test somehow to do this - I was trying to avoid being too fancy with these tests, but looks like I'll need to rethink them based on the comments so far.

ci/L0_multi_gpu/vllm_backend/test.sh
rmccorm4: There's only a single test right now, just lots of helpers. If I move the server/model setup into the python test like you mentioned, then the bash part can be simplified.

ci/L0_multi_gpu/vllm_backend/test.sh
rmccorm4: Do you or @jbkyang-nvi have a reference for that? If not, I can probably just do all this stuff inside the pytest using the in-process python API a bit more easily if we don't need any frontend features.

GuanLuo: It's just turn what we do in bash to in python (spawn process / file system manipulation etc.)

Seems like the change is reverted, sad.
https://github.com/triton-inference-server/server/pull/7195#discussion_r1600824526

rmccorm4: I agree a common set of utils to prep/start/stop server via python+subprocess would be great. That would probably take me some time to write something good though. Can I merge these tests using bash and follow-up with this after we deal with the P0's and pipeline failures? I'll take this test as a specific example to refactor using the common util I write. @GuanLuo @oandreeva-nv

GuanLuo (approved):

oandreeva-nv (approved):

rmccorm4:
Follow-up ticket: DLIS-6804

PR #45, opened 2024-07-03
Removed explicit mode for multi-lora

oandreeva-nv:
explicit model load was causing failures in tests with vllm==v0.5.0.post1

### DIFF
--- a/ci/L0_multi_gpu/multi_lora/multi_lora_test.py
+++ b/ci/L0_multi_gpu/multi_lora/multi_lora_test.py
@@ -119,7 +119,6 @@ class VLLMTritonLoraTest(AsyncTestResultCollector):
         self.triton_client.stop_stream()
 
     def test_multi_lora_requests(self):
-        self.triton_client.load_model(self.vllm_model_name)
         sampling_parameters = {"temperature": "0", "top_p": "1"}
         # make two requests separately to avoid the different arrival of response answers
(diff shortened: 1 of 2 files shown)


ci/L0_multi_gpu/multi_lora/multi_lora_test.py
```diff
@@ -119,7 +119,6 @@ def _test_vllm_model(
         self.triton_client.stop_stream()
 
     def test_multi_lora_requests(self):
-        self.triton_client.load_model(self.vllm_model_name)
```
Tabrizian, line 122: Do we know why it doesn't work with explicit mode?

oandreeva-nv: Due to the way vllm evaluates free GPU memory. I'm not sure what exactly changed, but in explicit mode `initialize` fails at:
```
 Error in memory profiling. This happens when the GPU memory was not properly cleaned up before initializing the vLLM instance.
 ```
Unfortunately, I don't have the capacity to investigate what exactly changed in the profiling

tanmayv25: @oandreeva-nv  Have you created ticket for further investigation?

oandreeva-nv: DLIS-7012

tanmayv25 (approved):

PR #47, opened 2024-07-17
test: Limiting multi-gpu tests to use Ray as distributed_executor_backend 

oandreeva-nv:
In PR [#5230](https://github.com/vllm-project/vllm/pull/5230) vllm changed default executor for distributed serving from Ray to `python native multiprocessing` for single node processing. This becomes in issue to Triton starting with v0.5.1 release.
For `python native multiprocessing` mode and KIND_MODEL setting, Triton hits "failed to stop server: Internal - Exit timeout expired. Exiting immediately." and `pt_main_thread` processes are never stopped/killed. I'll create an issue a bit later.

Solution: support only Ray for deploying models with tensor_parallel_size > 1 via "distributed_executor_backend" flag until the issue is fixed. 

This PR adjusts our multi-gpu tests according to the above observations

### DIFF
--- a/ci/L0_multi_gpu/multi_lora/test.sh
+++ b/ci/L0_multi_gpu/multi_lora/test.sh
@@ -62,7 +62,8 @@ model_json=$(cat <<EOF
     "enforce_eager": "true",
     "enable_lora": "true",
     "max_lora_rank": 32,
-    "lora_extra_vocab_size": 256
+    "lora_extra_vocab_size": 256,
+    "distributed_executor_backend":"ray"
 }
 EOF
 )
@@ -120,7 +121,8 @@ model_json=$(cat <<EOF
     "block_size": 16,
     "enforce_eager": "true",
     "enable_lora": "false",
-    "lora_extra_vocab_size": 256
+    "lora_extra_vocab_size": 256,
+    "distributed_executor_backend":"ray"
 }
 EOF
 )
--- a/ci/L0_multi_gpu/vllm_backend/test.sh
+++ b/ci/L0_multi_gpu/vllm_backend/test.sh
@@ -30,7 +30,7 @@ source ../../common/util.sh
 TRITON_DIR=${TRITON_DIR:="/opt/tritonserver"}
 SERVER=${TRITON_DIR}/bin/tritonserver
 BACKEND_DIR=${TRITON_DIR}/backends
-SERVER_ARGS="--model-repository=`pwd`/models --backend-directory=${BACKEND_DIR} --model-control-mode=explicit --log-verbose=1"
+SERVER_ARGS="--model-repository=`pwd`/models --backend-directory=${BACKEND_DIR} --log-verbose=1"
 TEST_RESULT_FILE='test_results.txt'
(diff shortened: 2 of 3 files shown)


ci/L0_multi_gpu/multi_lora/test.sh
```diff
@@ -62,7 +62,8 @@ model_json=$(cat <<EOF
     "enforce_eager": "true",
     "enable_lora": "true",
     "max_lora_rank": 32,
-    "lora_extra_vocab_size": 256
+    "lora_extra_vocab_size": 256,
+    "distributed_executor_backend":"ray"
```
rmccorm4, line 66: > For python native multiprocessing mode and KIND_MODEL setting, Triton hits "failed to stop server: Internal - Exit timeout expired.

Few questions on this:
1. Do we know more details or limitations on why this is happening?
2. Is this an error happening on server shutdown?
    - Is there some issue with the `python native multiprocessing` due to the details of Triton's python backend launching each instance as a separate process? 
3. Is this with 1, 2, or any amount of model instances with KIND_MODEL?

rcarrata:
@oandreeva-nv afaik you can setup the --distributed-executor-backend to ray and avoid the usage of MP.

From the [docs of distributed serving](https://docs.vllm.ai/en/stable/serving/distributed_serving.html):
`Multiprocessing will be used by default when not running in a Ray placement group and if there are sufficient GPUs available on the same node for the configured tensor_parallel_size, otherwise Ray will be used. This default can be overridden via the LLM class distributed-executor-backend argument or --distributed-executor-backend API server argument. Set it to mp for multiprocessing or ray for Ray. It’s not required for Ray to be installed for the multiprocessing case.`

oandreeva-nv:
@rcarrata That's what I'm doing in this PR. I'm making sure that ray is used for distributed testing. Or did I misunderstood your comment?

ci/L0_multi_gpu/multi_lora/test.sh
oandreeva-nv: > Do we know more details or limitations on why this is happening?

Issue with unclear multi-gpu test fail when upgrading to 0.5.0 versions and up
In PR [#5230](https://github.com/vllm-project/vllm/pull/5230) vllm changed default executor for distributed serving from Ray to python native multiprocessing for single node processing. This becomes in issue starting with v0.5.1 release.
For `python native multiprocessing` mode and KIND_MODEL setting, triton hits "failed to stop server: Internal - Exit timeout expired. Exiting immediately." and pt_main_thread processes are never stopped/killed.
Solution: add `"distributed_executor_backend":"ray"`  to `model.json`

> Is this an error happening on server shutdown? 

Yes, and I have a reproducer outside of triton

> Is this with 1, 2, or any amount of model instances with KIND_MODEL?

If "distributed_executor_backend" field is not specified, than for tp>2 and distributed among a single node, than `MP` backend kicks in. However, I've noticed that even when tp=1 and "distributed_executor_backend" is specified in `model.json`, vllm will go through distributed serving even when tp=1. More on the slack channel for this behavior

fpetrini15 (approved):

PR #46, opened 2024-07-10
perf: Improve vLLM backend performance by using a separate thread for responses

Tabrizian:
#### What does the PR do?

Triton's output token throughput for generate endpoint increases by 18% for concurrency 50. There is still a small gap between vLLM-only and vLLM + Triton solution.

The model is llama-2-7b.

Changes:
* Use add_request instead of generate for vLLM backend. It appears that with generate , the token generation is delayed until the next iteration of the loop.
* Delegate response sending to a separate thread.

Next steps:
* In the long term it might be better to have an async_send API to avoid creating a separate thread for sending the responses.

image

image

#### Checklist
- [x] PR title reflects the change and is of format `<commit_type>: <Title>`
- [x] Changes are described in the pull request.
- [x] Related issues are referenced.
- [ ] Populated [github labels](https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/managing-labels) field
- [ ] Added [test plan](#test-plan) and verified test passes.
- [ ] Verified that the PR passes existing CI.
- [x] Verified copyright is correct on all changed files.
- [ ] Added _succinct_ git squash message before merging [ref](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html).
- [ ] All template sections are filled out.
- [ ] Optional: Additional screenshots for behavior/output changes with before/after.

#### Commit Type:
Check the [conventional commit type](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type)
box here and add the label to the github PR.
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [x] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Related PRs:

N/A

#### Where should the reviewer start?

N/A

#### Test plan:

This is a performance improvement, existing test cases should be sufficient at covering any possible issues.

- CI Pipeline ID: 16917857 16926866

#### Caveats:

N/A

#### Background

N/A

#### Related Issues: (use one of the action keywords Closes / Fixes / Resolves / Relates to)
N/A

PR #48, opened 2024-07-22
fix: Fix non garbage collect on response sender

kthui:
#### What does the PR do?

Response sender(s) may not be deleted and garbage collected properly when it is passed to another thread. This PR ensures it is deleted when it goes out of scope.

#### Checklist
- [ ] Added [test plan](#test-plan) and verified test passes.
- [ ] Verified that the PR passes existing CI.

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [x] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Related PRs:

https://github.com/triton-inference-server/vllm_backend/pull/46

Start with the [related PR](https://github.com/triton-inference-server/vllm_backend/pull/46) and then look into this fix and some minor refactoring.

#### Test plan:

Existing tests on vLLM backend should be sufficient at covering any regression bring by the performance improvement (and fix).

- CI Pipeline ID:

#### Caveats:

N/A

#### Background

N/A

- closes GitHub issue: N/A

### DIFF
--- a/src/model.py
+++ b/src/model.py
@@ -25,6 +25,7 @@
 # OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 import asyncio
+import gc
 import json
 import os
 import queue
@@ -114,39 +115,18 @@ class TritonPythonModel:
         # Counter to keep track of ongoing request counts
         self.ongoing_request_count = 0
 
+        # Starting the response thread
+        self._response_queue = queue.Queue()
+        self._response_thread = threading.Thread(target=self.response_loop)
+        self._response_thread.start()
+
         # Starting asyncio event loop to process the received requests asynchronously.
         self._loop = asyncio.get_event_loop()
         self._loop_thread = threading.Thread(
             target=self.engine_loop, args=(self._loop,)
         )
         self._shutdown_event = asyncio.Event()
-        self._response_queue = queue.Queue()
-        self._thread = threading.Thread(target=self.send_responses)
-        self._thread.start()
         self._loop_thread.start()
-
-    def send_responses(self):
-        while True:
-            item = self._response_queue.get()
-            print(item)
-            # To signal shutdown a None item will be added to the queue.
-            if item is None:
-                break
-            response_sender = item[0]
-            response = item[1]
-            response_flag = item[2]
-            try:
-                response_sender.send(response, response_flag)
-            except Exception as e:
-                self.logger.log_error(
-                    f"An error occurred while sending a response: {e}"
-                )
-            finally:
-                if response_flag == pb_utils.TRITONSERVER_RESPONSE_COMPLETE_FINAL:
-                    self.ongoing_request_count -= 1
-                response_sender = None
-                response = None
-                response_flag = None
 
     def init_engine(self):
         # Currently, Triton needs to use decoupled policy for asynchronously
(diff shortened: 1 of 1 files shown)


src/model.py
```diff
@@ -300,6 +280,28 @@ def get_sampling_params_dict(self, params_json):
...
+            except Exception as e:
+                self.logger.log_error(
+                    f"An error occurred while sending a response: {e}"
+                )
+            finally:
+                if response_flag == pb_utils.TRITONSERVER_RESPONSE_COMPLETE_FINAL:
+                    self.ongoing_request_count -= 1
+                    del response_sender
+                    if self._response_queue.empty():
+                        gc.collect()
```
Tabrizian, line 302: Can we do garbage collection on `None` item only? I'm a bit concerned with calling garbage collection for every request completion.

kthui: > Can we do garbage collection on `None` item only?

The `None` item is only placed into the response queue when the `finalize()` function is called, but the `finalize()` function is called only if all the response sender(s) are destructed. In other words, we will only see the `None` item after the `response_sender`(s) are garbage collected.

I think the best solution now is to garbage collect if the response queue is empty, which means garbage collect happens only if there is no new response to be sent at the moment a request is completed, so it should not create too much overhead ideally.

Here is the frequency and duration of garbage collects happened during benchmarking:
```
I0722 21:40:56.330660 2595 grpc_server.cc:2463] "Started GRPCInferenceService at 0.0.0.0:8001"
I0722 21:40:56.330826 2595 http_server.cc:4692] "Started HTTPService at 0.0.0.0:8000"
I0722 21:40:56.383418 2595 http_server.cc:362] "Started Metrics Service at 0.0.0.0:8002"
I0722 21:41:30.977492 2595 model.py:302] "----- Garabage collect start -----"
I0722 21:41:31.072659 2595 model.py:304] "----- Garabage collect end -----"
I0722 21:41:55.523516 2595 model.py:302] "----- Garabage collect start -----"
I0722 21:41:55.618196 2595 model.py:304] "----- Garabage collect end -----"
I0722 21:42:20.098470 2595 model.py:302] "----- Garabage collect start -----"
I0722 21:42:20.193094 2595 model.py:304] "----- Garabage collect end -----"
I0722 21:42:44.672155 2595 model.py:302] "----- Garabage collect start -----"
I0722 21:42:44.766406 2595 model.py:304] "----- Garabage collect end -----"
I0722 21:43:09.278387 2595 model.py:302] "----- Garabage collect start -----"
I0722 21:43:09.372832 2595 model.py:304] "----- Garabage collect end -----"
I0722 21:43:33.794605 2595 model.py:302] "----- Garabage collect start -----"
I0722 21:43:33.889078 2595 model.py:304] "----- Garabage collect end -----"
I0722 21:43:58.329119 2595 model.py:302] "----- Garabage collect start -----"
I0722 21:43:58.422977 2595 model.py:304] "----- Garabage collect end -----"
I0722 21:44:22.812579 2595 model.py:302] "----- Garabage collect start -----"
I0722 21:44:22.906684 2595 model.py:304] "----- Garabage collect end -----"
```

Tabrizian (approved):
We need to fix the lifetime issue of finalize not getting called when the response factory is not deleted in the core.

GuanLuo:
Why response sender needs to be gc'ed right the way?

kthui:
> Why response sender needs to be gc'ed right the way?

Because garbage collecting the response sender will call its destructor on the C++ side, which then calls the destructor on the response factory which holds a shared pointer to the model. The shared pointer will stop the model from gracefully unloading.

We are not garbage collecting right away, it will only collect based on a heuristic that determines if there is no more ongoing request, which generally only collect once at the end of a busy benchmarking.

PR #53, opened 2024-07-29
feat: Add vLLM counter metrics access through Triton

yinggeh:
Sample endpoint output
```
# HELP vllm:prompt_tokens_total Number of prefill tokens processed.
# TYPE vllm:prompt_tokens_total counter
vllm:prompt_tokens_total{model="vllm_metrics",version="1"} 60
# HELP vllm:generation_tokens_total Number of generation tokens processed.
# TYPE vllm:generation_tokens_total counter
vllm:generation_tokens_total{model="vllm_metrics",version="1"} 96
```

#### What does the PR do?
Add vLLM counter metrics access through python_backend custom metrics.

#### Checklist
- [x] Added _succinct_ git squash message before merging [ref](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html).

#### Commit Type:
- [x] feat

#### Related PRs:
https://github.com/triton-inference-server/server/pull/7493

n/a

#### Test plan:
L0_backend_vllm/metrics_test

- CI Pipeline ID:
17372863

#### Caveats:
1. Update user guide and docs

#### Background
Customers requested.

n/a


src/metrics.py
```diff
@@ -0,0 +1,286 @@
...
+            kind=pb_utils.MetricFamily.COUNTER)
+        self.counter_prompt_tokens_family = pb_utils.MetricFamily(
+            name="vllm:prompt_tokens_total",
+            description="Number of prefill tokens processed.",
+            kind=pb_utils.MetricFamily.COUNTER)
+        self.counter_generation_tokens_family = pb_utils.MetricFamily(
+            name="vllm:generation_tokens_total",
+            description="Number of generation tokens processed.",
+            kind=pb_utils.MetricFamily.COUNTER)
+        # self.histogram_time_to_first_token_family = pb_utils.MetricFamily(
```
yinggeh, line 74: Every commented metric is unsupported at this point. Will be cleaned before merged.

src/model.py
```diff
@@ -151,6 +153,11 @@ def init_engine(self):
             AsyncEngineArgs(**self.vllm_engine_config)
         )
 
+        # Create vLLM custom Metrics
+        labels = {"model": "vllm_metrics", "version": "1"}
```
rmccorm4, line 157: I don't think this should be hard-coded, it should match the Triton model name/version:
```python
labels = {"model": self.args["model_name"], "version": self.args["model_version"]}
```
```suggestion
        labels = {"model": self.args["model_name"], "version": self.args["model_version"]}
```

src/model.py
```diff
@@ -151,6 +153,11 @@ def init_engine(self):
...
+        logger = VllmStatLogger(vllm_labels=labels)
```
rmccorm4, line 158: I think the `vllm_labels` arg name is wrong

src/metrics.py
```diff
@@ -0,0 +1,301 @@
...
+        # self.counter_spec_decode_num_draft_tokens = self.counter_spec_decode_num_draft_tokens_family.Metric(
+        # labels=labels
+        # )
+        # self.counter_spec_decode_num_emitted_tokens = self.counter_spec_decode_num_emitted_tokens_family.Metric(
+        #     labels=labels
+        # )
+
+
+class VllmStatLogger(VllmStatLoggerBase):
+    """StatLoggeris used as adapter between vLLM stats collector and Triton metrics provider."""
```
rmccorm4, line 237:
```suggestion
    """StatLogger is used as an adapter between vLLM stats collector and Triton metrics provider."""
```

GuanLuo:
test?

src/model.py
yinggeh: Fixed

src/model.py
yinggeh: Fixed

src/metrics.py
yinggeh: Fixed

src/model.py
```diff
@@ -151,6 +153,14 @@ def init_engine(self):
             AsyncEngineArgs(**self.vllm_engine_config)
         )
 
+        # Create vLLM custom Metrics
+        labels = {
+            "model": self.args["model_name"],
+            "version": self.args["model_version"],
+        }
+        logger = VllmStatLogger(labels=labels)
+        self.llm_engine.add_logger("triton", logger)
```
rmccorm4, line 162: What is the cadence that the logger gets called? CC @kthui as this will involve round trips with core, similar to your investigation with request cancellation frequency.

yinggeh: Can you elaborate on this?

rmccorm4: How often will the metrics get updated? Every request, every token, every full response, etc. ? in other words, how often will vLLM engine call this attached triton stats logger?

GuanLuo: every iteration

rmccorm4: That will probably significantly affect the total throughput then if the core round trip communication will interrupt the generation at every iteration based on Jacky and Iman's recent findings. We probably want this feature either way - just calling out that we'll likely need to make similar optimizations that @kthui is doing right now for this feature.

src/metrics.py
```diff
@@ -0,0 +1,301 @@
...
+
+import triton_python_backend_utils as pb_utils
+from vllm.engine.metrics import StatLoggerBase as VllmStatLoggerBase
+from vllm.engine.metrics import Stats as VllmStats
+from vllm.engine.metrics import SupportsMetricsInfo
+
+
+# begin-metrics-definitions
+class TritonMetrics:
+    def __init__(self, labels):
```
rmccorm4, line 37: Do we need to verify if server metrics are enabled? I think if you build server with `TRITON_ENABLE_METRICS=OFF`, then the vLLM backend will become unusable unless you put some safeguards to only use metrics if enabled.

src/metrics.py
```diff
@@ -0,0 +1,301 @@
...
+
+from typing import Dict, Union
+
+import triton_python_backend_utils as pb_utils
+from vllm.engine.metrics import StatLoggerBase as VllmStatLoggerBase
+from vllm.engine.metrics import Stats as VllmStats
+from vllm.engine.metrics import SupportsMetricsInfo
+
+
+# begin-metrics-definitions
```
oandreeva-nv, line 35:
```suggestion
```

oandreeva-nv:
Do we have a corresponding PR on the server side? Right now during the build container only copies `model.py` from this github repo, if we're adding `metrics.py`, that should be copied as well. Ideally, I suggest putting `metics.py` under `utils/` folder and add logic to `build.py` that copies `utils` folder as well

oandreeva-nv:
I would like to have a look at this PR as well, before it gets merged

yinggeh:
> Do we have a corresponding PR on the server side? Right now during the build container only copies `model.py` from this github repo, if we're adding `metrics.py`, that should be copied as well. Ideally, I suggest putting `metics.py` under `utils/` folder and add logic to `build.py` that copies `utils` folder as well
Updated. https://github.com/triton-inference-server/server/pull/7493

ci/L0_backend_vllm/metrics_test/test.sh
```diff
@@ -0,0 +1,98 @@
+#!/bin/bash
+# Copyright 2023-2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
oandreeva-nv, line 2:
```suggestion
# Copyright 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

ci/L0_backend_vllm/metrics_test/vllm_metrics_test.py
```diff
@@ -0,0 +1,171 @@
+# Copyright 2023-2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
oandreeva-nv, line 1:
```suggestion
# Copyright 2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

src/utils/metrics.py
```diff
@@ -0,0 +1,156 @@
...
+
+import triton_python_backend_utils as pb_utils
+from vllm.engine.metrics import StatLoggerBase as VllmStatLoggerBase
+from vllm.engine.metrics import Stats as VllmStats
+from vllm.engine.metrics import SupportsMetricsInfo
+
+
+class TritonMetrics:
+    def __init__(self, labels):
+        # System stats
```
oandreeva-nv, line 37: Let's make comments more descriptive, e.g. L54 and L94 are identical, which makes a wrong impression of a duplicated code.

src/utils/metrics.py
```diff
@@ -0,0 +1,156 @@
...
+    def __init__(self, labels: Dict, local_interval: float = 0) -> None:
+        # Tracked stats over current local logging interval.
+        super().__init__(local_interval)
+        self.metrics = TritonMetrics(labels=labels)
+
+    def info(self, type: str, obj: SupportsMetricsInfo) -> None:
+        raise NotImplementedError
+
+    def _log_gauge(self, gauge, data: Union[int, float]) -> None:
+        # Convenience function for logging to gauge.
```
oandreeva-nv, line 127: How about we convert the description to a Google docstring format : https://github.com/google/styleguide/blob/gh-pages/pyguide.md#383-functions-and-methods .

ci/L0_backend_vllm/metrics_test/vllm_metrics_test.py
```diff
@@ -0,0 +1,171 @@
...
+            prompts=PROMPTS,
+            sampling_parameters=SAMPLING_PARAMETERS,
+            stream=False,
+            send_parameters_as_tensor=True,
+            model_name=self.vllm_model_name,
+        )
+        expected_metrics_dict["vllm:prompt_tokens_total"] = 18
+        expected_metrics_dict["vllm:generation_tokens_total"] = 48
+        print(get_metrics())
+        print(expected_metrics_dict)
```
oandreeva-nv, line 100: do we need those prints?

ci/L0_backend_vllm/metrics_test/vllm_metrics_test.py
```diff
@@ -0,0 +1,171 @@
...
+
+PROMPTS = [
+    "The most dangerous animal is",
+    "The capital of France is",
+    "The future of AI is",
+]
+SAMPLING_PARAMETERS = {"temperature": "0", "top_p": "1"}
+
+
+def get_metrics():
```
oandreeva-nv, line 50: Can we move this into `VLLMTritonMetricsTest` as a helper function?

ci/L0_backend_vllm/metrics_test/vllm_metrics_test.py
```diff
@@ -0,0 +1,171 @@
...
+            send_parameters_as_tensor=True,
+            model_name=self.vllm_model_name,
+        )
+        expected_metrics_dict["vllm:prompt_tokens_total"] = 18
+        expected_metrics_dict["vllm:generation_tokens_total"] = 48
+        print(get_metrics())
+        print(expected_metrics_dict)
+        self.assertEqual(get_metrics(), expected_metrics_dict)
+
+        self._test_vllm_model(
```
oandreeva-nv, line 103: Do we need this test or we just need to send a single request? My concern is if the `_test_vllm_model` is identical to [this](https://github.com/triton-inference-server/vllm_backend/blob/128abc3082d6c7963f5015ddeb6cffb09675799b/ci/L0_backend_vllm/vllm_backend/vllm_backend_test.py#L166) , so we would need to maintain it in both places with this PR

ci/L0_backend_vllm/metrics_test/vllm_metrics_test.py
```diff
@@ -0,0 +1,171 @@
...
+            prompts=PROMPTS,
+            sampling_parameters=SAMPLING_PARAMETERS,
+            stream=False,
+            send_parameters_as_tensor=False,
+            model_name=self.vllm_model_name,
+        )
+        expected_metrics_dict["vllm:prompt_tokens_total"] = 36
+        expected_metrics_dict["vllm:generation_tokens_total"] = 96
+        self.assertEqual(get_metrics(), expected_metrics_dict)
```
oandreeva-nv, line 112: So we only check 2 metrics, what about others?

ci/L0_backend_vllm/metrics_test/vllm_metrics_test.py
yinggeh: Sorry those were for debugging.

ci/L0_backend_vllm/metrics_test/vllm_metrics_test.py
```diff
@@ -0,0 +1,171 @@
...
+
+def get_metrics():
+    """
+    Store vllm metrics in a dictionary.
+    """
+    r = requests.get(f"http://{_tritonserver_ipaddr}:8002/metrics")
+    r.raise_for_status()
+
+    # Regular expression to match the pattern
+    pattern = r"^(vllm:.*){.*} (\d+)$"
```
rmccorm4, line 58: Does this regex work on the percentage metrics containing a decimal point `.` like `vllm:gpu_cache_usage_perc 0.5` ?

ci/L0_backend_vllm/metrics_test/vllm_metrics_test.py
```diff
@@ -0,0 +1,171 @@
...
+            sampling_parameters=SAMPLING_PARAMETERS,
+            stream=False,
+            send_parameters_as_tensor=True,
+            model_name=self.vllm_model_name,
+        )
+        expected_metrics_dict["vllm:prompt_tokens_total"] = 18
+        expected_metrics_dict["vllm:generation_tokens_total"] = 48
+        print(get_metrics())
+        print(expected_metrics_dict)
+        self.assertEqual(get_metrics(), expected_metrics_dict)
```
rmccorm4, line 101: Is this test passing? I'm curious if the cache percentage metrics are getting correctly matched.

ci/L0_backend_vllm/metrics_test/vllm_metrics_test.py
```diff
@@ -0,0 +1,171 @@
...
+            self.assertIsNotNone(output, "`text_output` should not be None")
+            if expected_output is not None:
+                self.assertEqual(
+                    output,
+                    expected_output[i],
+                    'Actual and expected outputs do not match.\n \
+                                  Expected "{}" \n Actual:"{}"'.format(
+                        output, expected_output[i]
+                    ),
+                )
```
rmccorm4, line 162: Are we using `expected_output` in this test?

src/utils/metrics.py
yinggeh: Updated

ci/L0_backend_vllm/metrics_test/vllm_metrics_test.py
yinggeh: Updated

ci/L0_backend_vllm/metrics_test/vllm_metrics_test.py
yinggeh: Updated

ci/L0_backend_vllm/metrics_test/vllm_metrics_test.py
yinggeh: Updated

ci/L0_backend_vllm/metrics_test/vllm_metrics_test.py
yinggeh: No. But all metrics results from vllm are integers. For KV-cache usages, 1 means 100 percent usage.

rmccorm4: If 1 means 100%, then isn't 0.5 == 50%?

ci/L0_backend_vllm/metrics_test/vllm_metrics_test.py
yinggeh: Yes. All metrics were zeros except "vllm:prompt_tokens_total" and "vllm:generation_tokens_total".

src/utils/metrics.py
yinggeh: Updated

ci/L0_backend_vllm/metrics_test/vllm_metrics_test.py
rmccorm4: Should they all be zero?

ci/L0_backend_vllm/metrics_test/vllm_metrics_test.py
rmccorm4: They're defined as `float` as far as I can tell: https://github.com/vllm-project/vllm/blob/8d59dbb00044a588cab96bcdc028006ed922eb06/vllm/engine/metrics.py#L309-L310

yinggeh: I removed most metrics and only kept "vllm:prompt_tokens_total" and "vllm:generation_tokens_total" based on customer's request. I will fix the regex. Don't worry about decimal point for now.

ci/L0_backend_vllm/metrics_test/vllm_metrics_test.py
yinggeh: Values in stats during one request.
```
num_running_sys: 1
num_waiting_sys: 0
num_swapped_sys: 0
gpu_cache_usage_sys: 2.3890295761885838e-05
cpu_cache_usage_sys: 0.0
num_preemption_iter: 0
num_prompt_tokens_iter: 8
num_generation_tokens_iter: 1.0

num_running_sys: 1
num_waiting_sys: 0
num_swapped_sys: 0
gpu_cache_usage_sys: 2.3890295761885838e-05
cpu_cache_usage_sys: 0.0
num_preemption_iter: 0
num_prompt_tokens_iter: 0
num_generation_tokens_iter: 1.0

num_running_sys: 1
num_waiting_sys: 0
num_swapped_sys: 0
gpu_cache_usage_sys: 2.3890295761885838e-05
cpu_cache_usage_sys: 0.0
num_preemption_iter: 0
num_prompt_tokens_iter: 0
num_generation_tokens_iter: 1.0

num_running_sys: 1
num_waiting_sys: 0
num_swapped_sys: 0
gpu_cache_usage_sys: 2.3890295761885838e-05
cpu_cache_usage_sys: 0.0
num_preemption_iter: 0
num_prompt_tokens_iter: 0
num_generation_tokens_iter: 1.0

num_running_sys: 1
num_waiting_sys: 0
num_swapped_sys: 0
gpu_cache_usage_sys: 2.3890295761885838e-05
cpu_cache_usage_sys: 0.0
num_preemption_iter: 0
num_prompt_tokens_iter: 0
num_generation_tokens_iter: 1.0

num_running_sys: 1
num_waiting_sys: 0
num_swapped_sys: 0
gpu_cache_usage_sys: 2.3890295761885838e-05
cpu_cache_usage_sys: 0.0
num_preemption_iter: 0
num_prompt_tokens_iter: 0
num_generation_tokens_iter: 1.0

num_running_sys: 1
num_waiting_sys: 0
num_swapped_sys: 0
gpu_cache_usage_sys: 2.3890295761885838e-05
cpu_cache_usage_sys: 0.0
num_preemption_iter: 0
num_prompt_tokens_iter: 0
num_generation_tokens_iter: 1.0

num_running_sys: 1
num_waiting_sys: 0
num_swapped_sys: 0
gpu_cache_usage_sys: 2.3890295761885838e-05
cpu_cache_usage_sys: 0.0
num_preemption_iter: 0
num_prompt_tokens_iter: 0
num_generation_tokens_iter: 1.0

num_running_sys: 1
num_waiting_sys: 0
num_swapped_sys: 0
gpu_cache_usage_sys: 2.3890295761885838e-05
cpu_cache_usage_sys: 0.0
num_preemption_iter: 0
num_prompt_tokens_iter: 0
num_generation_tokens_iter: 1.0

num_running_sys: 1
num_waiting_sys: 0
num_swapped_sys: 0
gpu_cache_usage_sys: 4.7780591523771676e-05
cpu_cache_usage_sys: 0.0
num_preemption_iter: 0
num_prompt_tokens_iter: 0
num_generation_tokens_iter: 1.0

num_running_sys: 1
num_waiting_sys: 0
num_swapped_sys: 0
gpu_cache_usage_sys: 4.7780591523771676e-05
cpu_cache_usage_sys: 0.0
num_preemption_iter: 0
num_prompt_tokens_iter: 0
num_generation_tokens_iter: 1.0

num_running_sys: 1
num_waiting_sys: 0
num_swapped_sys: 0
gpu_cache_usage_sys: 4.7780591523771676e-05
cpu_cache_usage_sys: 0.0
num_preemption_iter: 0
num_prompt_tokens_iter: 0
num_generation_tokens_iter: 1.0

num_running_sys: 1
num_waiting_sys: 0
num_swapped_sys: 0
gpu_cache_usage_sys: 4.7780591523771676e-05
cpu_cache_usage_sys: 0.0
num_preemption_iter: 0
num_prompt_tokens_iter: 0
num_generation_tokens_iter: 1.0

num_running_sys: 1
num_waiting_sys: 0
num_swapped_sys: 0
gpu_cache_usage_sys: 4.7780591523771676e-05
cpu_cache_usage_sys: 0.0
num_preemption_iter: 0
num_prompt_tokens_iter: 0
num_generation_tokens_iter: 1.0

num_running_sys: 1
num_waiting_sys: 0
num_swapped_sys: 0
gpu_cache_usage_sys: 4.7780591523771676e-05
cpu_cache_usage_sys: 0.0
num_preemption_iter: 0
num_prompt_tokens_iter: 0
num_generation_tokens_iter: 1.0

num_running_sys: 0
num_waiting_sys: 0
num_swapped_sys: 0
gpu_cache_usage_sys: 0.0
cpu_cache_usage_sys: 0.0
num_preemption_iter: 0
num_prompt_tokens_iter: 0
num_generation_tokens_iter: 1.0
```

ci/L0_backend_vllm/metrics_test/vllm_metrics_test.py
yinggeh: Updated

ci/L0_backend_vllm/metrics_test/vllm_metrics_test.py
yinggeh: Updated

ci/L0_backend_vllm/metrics_test/vllm_metrics_test.py
yinggeh: Updated

src/model.py
oandreeva-nv, line 162: Cross posting here: Question for this and `histogram` PR as well. Would this cause trouble if tritonserver was built without metrics ?

rmccorm4: +1: https://github.com/triton-inference-server/vllm_backend/pull/53#discussion_r1697679955

src/metrics.py
yinggeh: @rmccorm4 Fix for (a).
- https://github.com/triton-inference-server/vllm_backend/pull/53/commits/979dc02bc979362f7f28a39b0cd75320fb905b88
- https://github.com/triton-inference-server/python_backend/pull/375

Screenshot 2024-08-07 at 3 19 20 PM

src/model.py
yinggeh: @kthui will run benchmarks.

src/model.py
```diff
@@ -151,6 +153,16 @@ def init_engine(self):
             AsyncEngineArgs(**self.vllm_engine_config)
         )
 
+        # If TRITON_ENABLE_METRICS<_CPU/_GPU> build flag is enabled.
+        if self.args["metrics_mode"] in ["all", "cpu", "gpu"]:
```
oandreeva-nv, line 157: I'm curious if instead of adding a new `config` value, can we re-use any of vllm's arguments as a proxy?

yinggeh: {
    "model_config": '{
        "name": "vllm_model",
        "platform": "",
        "backend": "vllm",
        "runtime": "model.py",
        "version_policy": {"latest": {"num_versions": 1}},
        "max_batch_size": 0,
        "input": [
            {
                "name": "text_input",
                "data_type": "TYPE_STRING",
                "format": "FORMAT_NONE",
                "dims": [1],
                "is_shape_tensor": false,
                "allow_ragged_batch": false,
                "optional": false,
                "is_non_linear_format_io": false,
            },
            {
                "name": "stream",
                "data_type": "TYPE_BOOL",
                "format": "FORMAT_NONE",
                "dims": [1],
                "is_shape_tensor": false,
                "allow_ragged_batch": false,
                "optional": true,
                "is_non_linear_format_io": false,
            },
            {
                "name": "sampling_parameters",
                "data_type": "TYPE_STRING",
                "format": "FORMAT_NONE",
                "dims": [1],
                "is_shape_tensor": false,
                "allow_ragged_batch": false,
                "optional": true,
                "is_non_linear_format_io": false,
            },
            {
                "name": "exclude_input_in_output",
                "data_type": "TYPE_BOOL",
                "format": "FORMAT_NONE",
                "dims": [1],
                "is_shape_tensor": false,
                "allow_ragged_batch": false,
                "optional": true,
                "is_non_linear_format_io": false,
            },
        ],
        "output": [
            {
                "name": "text_output",
                "data_type": "TYPE_STRING",
                "dims": [-1],
                "label_filename": "",
                "is_shape_tensor": false,
                "is_non_linear_format_io": false,
            }
        ],
        "batch_input": [],
        "batch_output": [],
        "optimization": {
            "priority": "PRIORITY_DEFAULT",
            "input_pinned_memory": {"enable": true},
            "output_pinned_memory": {"enable": true},
            "gather_kernel_buffer_threshold": 0,
            "eager_batching": false,
        },
        "instance_group": [
            {
                "name": "vllm_model_0",
                "kind": "KIND_MODEL",
                "count": 1,
                "gpus": [],
                "secondary_devices": [],
                "profile": [],
                "passive": false,
                "host_policy": "",
            }
        ],
        "default_model_filename": "",
        "cc_model_filenames": {},
        "metric_tags": {},
        "parameters": {},
        "model_warmup": [],
        "model_transaction_policy": {"decoupled": true},
    },
    "model_instance_kind": "MODEL",
    "model_instance_name": "vllm_model_0_0",
    "model_instance_device_id": "0",
    "model_repository": "/mnt/backend_repos/vllm_backend/samples/model_repository/vllm_model",
    "model_version": "1",
    "model_name": "vllm_model",
    "metrics_mode": "off",
}

rmccorm4, line 157: Instead of introducing this new `metrics_mode` [here](https://github.com/triton-inference-server/python_backend/pull/375/commits/6404f14e0b62816048b6e1272a2521b1780810fc#diff-f2bb64cbc9d4dc5d938ce965813b23178f53732e02e8c3a2afa707e287e92e57R567-R583), does it make sense to do something simpler like this?

```python
class VllmStatLogger:
  def __init__(self):
    try:
      self.metrics = TritonMetrics(labels=labels)
    # Get some "metrics not supported" error if metrics disabled
    except TritonModelException as e:
      self.metrics = None    

  def log(self):
    if not self.metrics:
      return

    # otherwise metrics are enabled so log_counter()
```

CC @Tabrizian @krishung5

rmccorm4: Or similarly only call `llm_engine.add_logger(logger)` if `logger` initializes successfully.

yinggeh: User might want to enable only CPU/GPU metrics from vLLM.

rmccorm4: Personally I think we should defer that fine-grained level of support which involves addition of a new `arg` to python models to further discussion, and not include it as part of this PR to try to keep the scope as simple as possible.

src/metrics.py
yinggeh: (b) I ran some inference and it looks normal to me. No crash no exception.

yinggeh: Metrics are still created in (b). Move the fix to the optimization ticket.

src/model.py
oandreeva-nv: These are Triton arguments, I mentioned vllm engine arguments: https://docs.vllm.ai/en/latest/models/engine_args.html .

oandreeva-nv: Speculating here, so needs to be verified. Based on quick look, I think `disable_log_stats` should control whether to collect stats or not, based on this: https://github.com/vllm-project/vllm/blob/21b9c49aa37c7ba08590a99b0d4f15f86439c8f9/vllm/engine/llm_engine.py#L450  and this: https://github.com/vllm-project/vllm/blob/21b9c49aa37c7ba08590a99b0d4f15f86439c8f9/vllm/engine/llm_engine.py#L319C17-L319C26

So I think we can read `disable_log_stats` from `model.json` and enable metrics based on the presence of this flag. CC @rmccorm4 ad @GuanLuo

oandreeva-nv: In any outcome of this discussion, we should also test the behavior of our backend when `disable_log_stats` is present in `model.json` and when it's not.

src/model.py
```diff
@@ -153,15 +167,9 @@ def init_engine(self):
...
-        if self.args["metrics_mode"] in ["all", "cpu", "gpu"]:
-            # Create vLLM custom Metrics
-            labels = {
-                "model": self.args["model_name"],
-                "version": self.args["model_version"],
-            }
-            logger = VllmStatLogger(labels=labels)
-            self.llm_engine.add_logger("triton", logger)
+        # Add vLLM custom metrics
+        if not self.metrics:
```
rmccorm4, line 171: I think this should be inverted, `if self.metrics: add_logger`

yinggeh: Yeah. Found that after testing.

samples/model_repository/vllm_model/1/model.json
yinggeh, line 1: Updating `model.json` due to parsing error.
```
AsyncEngineArgs(
    model="facebook/opt-125m",
    served_model_name=None,
    tokenizer="facebook/opt-125m",
    skip_tokenizer_init=False,
    tokenizer_mode="auto",
    trust_remote_code=False,
    download_dir=None,
    load_format="auto",
    dtype="auto",
    kv_cache_dtype="auto",
    quantization_param_path=None,
    seed=0,
    max_model_len=None,
    worker_use_ray=False,
    distributed_executor_backend=None,
    pipeline_parallel_size=1,
    tensor_parallel_size=1,
    max_parallel_loading_workers=None,
    block_size=16,
    enable_prefix_caching=False,
    disable_sliding_window=False,
    use_v2_block_manager=False,
    swap_space=4,
    cpu_offload_gb=0,
    gpu_memory_utilization=0.5,
    max_num_batched_tokens=None,
    max_num_seqs=256,
    max_logprobs=20,
    disable_log_stats="false",
    revision=None,
    code_revision=None,
    rope_scaling=None,
    rope_theta=None,
    tokenizer_revision=None,
    quantization=None,
    enforce_eager="true",
    max_context_len_to_capture=None,
    max_seq_len_to_capture=8192,
    disable_custom_all_reduce=False,
    tokenizer_pool_size=0,
    tokenizer_pool_type="ray",
    tokenizer_pool_extra_config=None,
    enable_lora=False,
    max_loras=1,
    max_lora_rank=16,
    enable_prompt_adapter=False,
    max_prompt_adapters=1,
    max_prompt_adapter_token=0,
    fully_sharded_loras=False,
    lora_extra_vocab_size=256,
    long_lora_scaling_factors=None,
    lora_dtype="auto",
    max_cpu_loras=None,
    device="auto",
    ray_workers_use_nsight=False,
    num_gpu_blocks_override=None,
    num_lookahead_slots=0,
    model_loader_extra_config=None,
    ignore_patterns=None,
    preemption_mode=None,
    scheduler_delay_factor=0.0,
    enable_chunked_prefill=None,
    guided_decoding_backend="outlines",
    speculative_model=None,
    speculative_draft_tensor_parallel_size=None,
    num_speculative_tokens=None,
    speculative_max_model_len=None,
    speculative_disable_by_batch_size=None,
    ngram_prompt_lookup_max=None,
    ngram_prompt_lookup_min=None,
    spec_decoding_acceptance_method="rejection_sampler",
    typical_acceptance_sampler_posterior_threshold=None,
    typical_acceptance_sampler_posterior_alpha=None,
    qlora_adapter_name_or_path=None,
    disable_logprobs_during_spec_decoding=None,
    otlp_traces_endpoint=None,
    engine_use_ray=False,
    disable_log_requests="true",
)
```

samples/model_repository/vllm_model/1/model.json
```diff
@@ -1,6 +1,7 @@
 {
     "model":"facebook/opt-125m",
-    "disable_log_requests": "true",
+    "disable_log_requests": true,
     "gpu_memory_utilization": 0.5,
-    "enforce_eager": "true"
+    "enforce_eager": true,
+    "disable_log_stats": false
```
yinggeh, line 6: We should add `disable_log_stats` to sample `model.json`. Open to suggestions.

ci/L0_backend_vllm/metrics_test/test.sh
```diff
@@ -0,0 +1,98 @@
...
+  if [ "$code" != "200" ]; then
+    cat ./curl.out
+    echo -e "\n***\n*** ${message} : line ${BASH_LINENO}\n***"
+    RET=1
+  fi
+}
+
+rm -rf models && mkdir -p models
+cp -r ${SAMPLE_MODELS_REPO}/vllm_model models/vllm_opt
+# `vllm_opt`` model will be loaded on server start and stay loaded throughout
```
oandreeva-nv, line 53:
```suggestion
# `vllm_opt` model will be loaded on server start and stay loaded throughout
```

ci/L0_backend_vllm/metrics_test/vllm_metrics_test.py
```diff
@@ -0,0 +1,136 @@
...
+                print(result.message())
+            self.assertIsNot(type(result), InferenceServerException, str(result))
+
+            output = result.as_numpy("text_output")
+            self.assertIsNotNone(output, "`text_output` should not be None")
+
+        self.triton_client.stop_stream()
+
+    def test_vllm_metrics(self):
+        # All vLLM metrics from tritonserver
```
oandreeva-nv, line 115: Clarify `All`. For anyone else, who would need to work on this test later, they should be fully aware of what this test tests. Something like: this test makes sure that counter metrics are collected properly.

BTW, how do we test `gauge` metrics?

samples/model_repository/vllm_model/1/model.json
oandreeva-nv: you can add this during test initialization

src/model.py
```diff
@@ -147,9 +149,28 @@ def init_engine(self):
...
 
         # Create an AsyncLLMEngine from the config from JSON
-        self.llm_engine = AsyncLLMEngine.from_engine_args(
-            AsyncEngineArgs(**self.vllm_engine_config)
-        )
+        aync_engine_args = AsyncEngineArgs(**self.vllm_engine_config)
+        self.llm_engine = AsyncLLMEngine.from_engine_args(aync_engine_args)
+
+        # Create vLLM custom metrics
+        if not aync_engine_args.disable_log_stats:
```
oandreeva-nv, line 156: We need to document this behavior

samples/model_repository/vllm_model/1/model.json
oandreeva-nv: parsing error when?

ci/L0_backend_vllm/metrics_test/test.sh
```diff
@@ -0,0 +1,160 @@
...
+    echo -e "\n***\n*** Failed to start $SERVER\n***"
+    exit 1
+fi
+
+set +e
+python3 $CLIENT_PY VLLMTritonMetricsTest.test_vllm_metrics_disabled -v > $CLIENT_LOG 2>&1
+
+if [ $? -ne 0 ]; then
+    cat $CLIENT_LOG
+    echo -e "\n***\n*** Running $CLIENT_PY FAILED. \n***"
```
krishung5, line 104: nit: Can we have the error message include the name of the test case so that it would be easier to check which test case fail on CI?

src/model.py
kthui: The current path forward is to allow metrics to be turned off. There is still room to improve in the future, i.e. perform the core round-trip communication on a side branch.

src/model.py
yinggeh: Resolved

ci/L0_backend_vllm/metrics_test/vllm_metrics_test.py
yinggeh: The test was updated to allow new metrics to be added.
We don't care about gauge metrics for now since there is none.
Most guage metrics have non-zero values in the middle of an inference but are set to zero when it finishes. We need to come up with a solution that captures all values have been set to gauge metrics.

src/model.py
yinggeh: Agree. I will update the doc in another PR after this gets merged due to time constraint.

ci/L0_backend_vllm/metrics_test/test.sh
yinggeh: Updated

ci/L0_backend_vllm/metrics_test/test.sh
yinggeh: Updated

samples/model_repository/vllm_model/1/model.json
yinggeh: Discussed offline. Adding `disable_log_stats` to sample config to let user aware of the change/metrics.

samples/model_repository/vllm_model/1/model.json
yinggeh: Discussed offline.

yinggeh: Discussed offline. So basically it worked because "true" is parsed as python `str` type and `bool("true") == True`. However,  `bool("false") == True` still so need this fix.

src/model.py
kthui: Do we need the documentation in 24.08? It needs to go in today if needed.

ci/L0_backend_vllm/metrics_test/vllm_metrics_test.py
oandreeva-nv: > We don't care about gauge metrics for now since there is none.
There is none, meaning?

src/utils/metrics.py
```diff
@@ -0,0 +1,111 @@
...
+    """StatLogger is used as an adapter between vLLM stats collector and Triton metrics provider."""
+
+    # local_interval not used here. It's for vLLM logs to stdout.
+    def __init__(self, labels: Dict, local_interval: float = 0) -> None:
+        # Tracked stats over current local logging interval.
+        super().__init__(local_interval)
+        self.metrics = TritonMetrics(labels=labels)
+
+    def info(self, type: str, obj: SupportsMetricsInfo) -> None:
+        raise NotImplementedError
```
GuanLuo, line 70: pass through.. You can't control if vLLM will call this function at some point in time and exception may not be handled..

src/model.py
yinggeh: I will follow up with another docs PR. Should be cherry-picked with histogram.

src/model.py
```diff
@@ -147,9 +149,28 @@ def init_engine(self):
...
+                if "metrics not supported" in str(e):
+                    # Metrics are disabled at the server
+                    self.metrics = None
+                    self.logger.log_info("[vllm] Metrics not supported")
+                else:
+                    raise e
+
+            # Add vLLM custom metrics
+            if self.metrics:
+                self.llm_engine.add_logger("triton", self.metrics)
```
GuanLuo, line 173: Why not move into `try` block

README.md
```diff
@@ -111,7 +111,9 @@ container with the following commands:
 
 ```
 mkdir -p /opt/tritonserver/backends/vllm
-wget -P /opt/tritonserver/backends/vllm https://raw.githubusercontent.com/triton-inference-server/vllm_backend/main/src/model.py
+git clone https://github.com/triton-inference-server/vllm_backend.git /opt/tritonserver/backends/vllm/vllm_backend
```
GuanLuo, line 114: Clone to a temporary directory instead of inside `/opt/tritonserver/backends`

src/model.py
kthui: At this point, the impact of having metrics (counter ***and gauge***) on performance with `--disable-log-stats` flag set on FastAPI completion vs Triton generate_stream is negligible. The delta between FastAPI completion and Triton generate_stream without any metrics functionality added is approximately the same with metrics and having the `--disable-log-stats` flag set.

src/model.py
yinggeh: OK. I was thinking to leave less call in a try catch statement.

README.md
yinggeh: `/tmp/backends`?

src/utils/metrics.py
yinggeh: Updated

src/model.py
yinggeh: Updated

README.md
yinggeh: Updated

README.md
```diff
@@ -110,8 +110,11 @@ In this case, please install vLLM first. You can do so by running
 container with the following commands:
 
 ```
-mkdir -p /opt/tritonserver/backends/vllm
-wget -P /opt/tritonserver/backends/vllm https://raw.githubusercontent.com/triton-inference-server/vllm_backend/main/src/model.py
+vllm_tmp_dir=/tmp/backends/vllm_backend
+mkdir -p /opt/tritonserver/backends/vllm $vllm_tmp_dir
+git clone https://github.com/triton-inference-server/vllm_backend.git $vllm_tmp_dir
+cp -r $vllm_tmp_dir/src/* /opt/tritonserver/backends/vllm
+rm -rf $vllm_tmp_dir
```
oandreeva-nv, line 117:
```suggestion
TEMP_TRITON_VLLM_DIR=/tmp/vllm_backend
mkdir -p /opt/tritonserver/backends/vllm $vllm_tmp_dir
git clone https://github.com/triton-inference-server/vllm_backend.git $TEMP_TRITON_VLLM_DIR
cp -r $TEMP_TRITON_VLLM_DIR/src/* /opt/tritonserver/backends/vllm
rm -rf $TEMP_TRITON_VLLM_DIR
```

oandreeva-nv: in current version you create `backends/vllm_backend` path, but delete only `vllm_backend` , `backends` folder is still left

src/utils/metrics.py
```diff
@@ -0,0 +1,99 @@
...
+            data: An int or float to increment the count metric.
+
+        Returns:
+            None
+        """
+        if data != 0:
+            counter.increment(data)
+
+    def log(self, stats: VllmStats) -> None:
+        """Logs tracked stats to triton metrics server every iteration.
```
oandreeva-nv, line 86: track ?

README.md
yinggeh: If `/tmp/backends` exists we shouldn't delete files other than vllm_backend in this directory.

yinggeh: Added `rm -d /tmp/backends` to prevent unexpected removal of files.

src/utils/metrics.py
yinggeh: Removed. Tracked stats was specific to vLLM's implementation of `PrometheusStatLogger`.

README.md
yinggeh: On second thought, I think it's fine to clone into `/tmp/vllm_backend`.

GuanLuo (dismissed):

src/utils/metrics.py
```diff
@@ -0,0 +1,99 @@
...
+            data: An int or float to increment the count metric.
+
+        Returns:
+            None
+        """
+        if data != 0:
+            counter.increment(data)
+
+    def log(self, stats: VllmStats) -> None:
+        """Logs to triton metrics server every iteration.
```
kthui, line 86: Can you either explain the meaning of "every iteration" on the comment, or simply don't mention it - for example: "Logs to triton metrics"?

src/utils/metrics.py
```diff
@@ -0,0 +1,99 @@
...
+import triton_python_backend_utils as pb_utils
+from vllm.engine.metrics import StatLoggerBase as VllmStatLoggerBase
+from vllm.engine.metrics import Stats as VllmStats
+from vllm.engine.metrics import SupportsMetricsInfo
+
+
+class TritonMetrics:
+    def __init__(self, labels):
+        # Initialize metric families
+        # Iteration stats
```
kthui, line 38: Can you elaborate on the meaning of "Iteration stats"?

kthui (dismissed):

src/utils/metrics.py
yinggeh: That's one of the vLLM metric catagories. See https://github.com/vllm-project/vllm/blob/fc93e5614374688bddc432279244ba7fbf8169c2/vllm/engine/metrics.py#L68

src/model.py
yinggeh: Docs added

README.md
```diff
@@ -194,14 +195,47 @@ starting from 23.10 release.
...
+# HELP vllm:generation_tokens_total Number of generation tokens processed.
+# TYPE vllm:generation_tokens_total counter
+vllm:generation_tokens_total{model="vllm_model",version="1"} 16
+```
+*Note:* The vLLM metrics reporting is disabled by default due to potential
+performance slowdowns. To enable vLLM model's metrics reporting, please add
+following lines to its config.pbtxt.
+```bash
+parameters: {
+  key: "REPORT_CUSTOM_METRICS"
```
oandreeva-nv, line 232: nit: not sure if it should be in caps

src/model.py
```diff
@@ -155,9 +157,28 @@ def init_engine(self):
...
-        )
+        aync_engine_args = AsyncEngineArgs(**self.vllm_engine_config)
+        self.llm_engine = AsyncLLMEngine.from_engine_args(aync_engine_args)
+
+        # Create vLLM custom metrics
+        if (
+            "REPORT_CUSTOM_METRICS" in self.model_config["parameters"]
+            and self.model_config["parameters"]["REPORT_CUSTOM_METRICS"]["string_value"]
+            == "yes"
+        ):
```
oandreeva-nv, line 168: potentially we can also check if `disable_log_stats` is `true`

README.md
yinggeh: To be consistent with the only `parameters` example found in our code
```
parameters: {
  key: "FORCE_CPU_ONLY_INPUT_TENSORS"
  value: {
    string_value:"no"
  }
}
```

oandreeva-nv: It's one example, here's lower case as well: https://github.com/triton-inference-server/server/blob/53200091b84f08a5e4921f5073137784570283e9/docs/user_guide/optimization.md#onnx-with-tensorrt-optimization-ort-trt

Tabrizian (approved):

Tabrizian:
LGTM, please make sure @oandreeva-nv's comments are addressed. Thanks Yingge!

src/utils/metrics.py
yinggeh: removed

README.md
yinggeh: I am more inclined to upper case for boolean value.

src/model.py
yinggeh: Nice catch. If `FORCE_CPU_ONLY_INPUT_TENSORS = true` but `disable_log_stats=true`, `add_logger()` throws exception. Test added.

oandreeva-nv: you mean `REPORT_CUSTOM_METRICS` not `FORCE_CPU_ONLY_INPUT_TENSORS` ?

oandreeva-nv (approved):
LGTM! Thanks for this work!

src/model.py
yinggeh: Sorry. Yes I meant `REPORT_CUSTOM_METRICS`.

README.md
kthui: Can use `key_value.upper()` before the comparison:
```
>>> "nO".upper()
'NO'
```

kthui (approved):
LGTM, nice work!

README.md
yinggeh: I'd rather make either all parameters case-insensitive at the time config.pbtxt was loaded or all case-sensitive.

PR #54, opened 2024-07-30
perf: Check for cancellation on response thread

kthui:
#### What does the PR do?

Cancellation check is to be performed during the response loop on the response thread. The status of the cancellation will be written back a object shared between the response loop and generate loop on a particular request. After sending a response, the cancellation status will be checked and the generate loop will cancel the generation if the request is cancelled. Combined with the release of GIL while checking for cancellation, this maximizes the vLLM generation time on the GIL, which has seen noticeable performance improvement.

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [x] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Related PRs:

https://github.com/triton-inference-server/python_backend/pull/372

Start at once a cancellation is issued while responses are being actively generated, does the generate loop capture the cancellation signal? Is the final flag sent (both streaming/non-streaming) after cancelling? Is the response sender deleted and garbage collected after cancelling?

#### Test plan:

This is a performance improvement, so any issue should be covered by existing test cases.

- CI Pipeline ID: 17061062

#### Caveats:

N/A

#### Background

Need to improve the performance on vLLM backend.

N/A

### DIFF
--- a/src/model.py
+++ b/src/model.py
@@ -424,7 +424,7 @@ class TritonPythonModel:
                         response_state["last_response_generated"] = True
                         response = pb_utils.InferenceResponse(
                             error=pb_utils.TritonError(
-                                message="cancelled on the client side",
+                                message="Request was cancelled",
                                 code=pb_utils.TritonError.CANCELLED,
                             )
                         )
--- a/src/model.py
+++ b/src/model.py
@@ -289,14 +289,12 @@ class TritonPythonModel:
                 break
             response_state, response, response_flag = item
             del item
-            response_sender, number_response_sent = response_state[0], response_state[3]
+            response_sender = response_state[0]
             try:
                 response_sender.send(response, response_flag)
-                response_state[3] = number_response_sent + 1
-                # Check for cancellation only if the last response is not yet generated
-                # and for every 10 response.
                 last_response_ready = response_state[2]
-                if not last_response_ready and number_response_sent % 2 == 0:
+                # Stop checking for cancellation if the last response is generated.
+                if not last_response_ready:
                     is_cancelled = response_sender.is_cancelled()
                     response_state[1] = is_cancelled
             except Exception as e:
@@ -355,7 +353,6 @@ class TritonPythonModel:
             response_sender,
             False,  # is cancelled
             False,  # last response ready to be sent
-            0,  # number of response sent
         ]
         self.ongoing_request_count += 1
         decrement_ongoing_request_count = True
--- a/src/model.py
+++ b/src/model.py
@@ -289,12 +289,14 @@ class TritonPythonModel:
                 break
             response_state, response, response_flag = item
             del item
-            response_sender = response_state[0]
+            response_sender, number_response_sent = response_state[0], response_state[3]
             try:
                 response_sender.send(response, response_flag)
+                response_state[3] = number_response_sent + 1
+                # Check for cancellation only if the last response is not yet generated
+                # and for every 10 response.
                 last_response_ready = response_state[2]
-                # Stop checking for cancellation if the last response is generated.
-                if not last_response_ready:
+                if not last_response_ready and number_response_sent % 2 == 0:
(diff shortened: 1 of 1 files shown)


src/model.py
```diff
@@ -342,6 +349,11 @@ async def generate(self, request):
         Forwards single request to LLM engine and returns responses.
         """
         response_sender = request.get_response_sender()
+        response_state = [
```
oandreeva-nv, line 352: maybe we can use dictionary instead of a list? this way in the code we can be more clear, i.e. instead of `response_state[1], we will use `response_state["is_cancelled"]`, or have a ResponseState class

src/model.py
```diff
@@ -403,10 +415,26 @@ async def generate(self, request):
...
+                if not stream:
+                    is_cancelled = response_sender.is_cancelled()
+                if is_cancelled:
                     self.logger.log_info("[vllm] Cancelling the request")
                     await self.llm_engine.abort(request_id)
                     self.logger.log_info("[vllm] Successfully cancelled the request")
+                    if stream:
+                        response_state[2] = True  # last response ready to be sent
+                        response = pb_utils.InferenceResponse(
+                            error=pb_utils.TritonError(
```
oandreeva-nv, line 428: Is this the logic we go by, i.e. if response is cancelled, we still send a response back with the message that it was cancelled?

src/model.py
kthui: yes, a dictionary is cleaner - let me do a quick check if there is any perf regression.

src/model.py
kthui: yes, for example: https://github.com/triton-inference-server/python_backend/blob/main/README.md#request-cancellation-handling

kthui: for decoupled, we also need to send complete final flag

oandreeva-nv: thanks! I would recommend changing `user cancelled` to `cancelled on the client side` or something like this, because it might be that request was cancelled by an intermediary system not related to user.

src/model.py
kthui: there is no regression, let's use the dictionary variant

src/model.py
kthui: sure, updated the message [Update cancellation message](https://github.com/triton-inference-server/vllm_backend/pull/54/commits/0b996e85c011ba26e928410b2039fe7b6bffdf3a)

src/model.py
```diff
@@ -403,10 +413,26 @@ async def generate(self, request):
             )
 
             async for output in response_iterator:
-                if response_sender.is_cancelled():
+                is_cancelled = response_state["is_cancelled"]
```
oandreeva-nv, line 416: I lack some understanding. Could you please clarify for non-stream case, when `response_state["is_cancelled"]` is updated/checked if the state was changed from `False` to `True` ?

oandreeva-nv: resolved offline

oandreeva-nv (dismissed):
LGTM! Please, let @Tabrizian to look at it as well

src/model.py
```diff
@@ -403,10 +413,26 @@ async def generate(self, request):
...
+                    is_cancelled = response_sender.is_cancelled()
+                if is_cancelled:
                     self.logger.log_info("[vllm] Cancelling the request")
                     await self.llm_engine.abort(request_id)
                     self.logger.log_info("[vllm] Successfully cancelled the request")
+                    if stream:
+                        response_state["last_response_generated"] = True
+                        response = pb_utils.InferenceResponse(
+                            error=pb_utils.TritonError(
+                                message="cancelled on the client side",
```
Tabrizian, line 427:
```suggestion
                                message="Request was cancelled",
```

Tabrizian (dismissed):
One minor comment, otherwise looks good.

Tabrizian (approved):

PR #61, opened 2024-08-20
feat: Report more histogram metrics

yinggeh:
#### What does the PR do?
Report more histogram metrics from vLLM to Triton metrics server.
```
# Histogram of end to end request latency in seconds.
histogram_e2e_time_request

# Number of prefill tokens processed.
histogram_num_prompt_tokens_request

# Number of generation tokens processed.
histogram_num_generation_tokens_request

# Histogram of the best_of request parameter.
histogram_best_of_request

# Histogram of the n request parameter.
histogram_n_request
```
#### Checklist

#### Commit Type:
- [x] feat

#### Related PRs:
n/a

n/a

#### Test plan:
ci/L0_backend_vllm/metrics_test

- CI Pipeline ID:
17681888

#### Caveats:
n/a

#### Background
Customers requested additional histogram metrics from vLLM.

n/a

### DIFF
--- a/ci/L0_backend_vllm/metrics_test/vllm_metrics_test.py
+++ b/ci/L0_backend_vllm/metrics_test/vllm_metrics_test.py
@@ -119,39 +119,60 @@ class VLLMTritonMetricsTest(TestResultCollector):
             model_name=self.vllm_model_name,
         )
         metrics_dict = self.parse_vllm_metrics()
+        total_prompts = len(self.prompts)
 
         # vllm:prompt_tokens_total
         self.assertEqual(metrics_dict["vllm:prompt_tokens_total"], 18)
         # vllm:generation_tokens_total
         self.assertEqual(metrics_dict["vllm:generation_tokens_total"], 48)
         # vllm:time_to_first_token_seconds
-        self.assertEqual(metrics_dict["vllm:time_to_first_token_seconds_count"], 3)
+        self.assertEqual(
+            metrics_dict["vllm:time_to_first_token_seconds_count"], total_prompts
+        )
         self.assertGreater(metrics_dict["vllm:time_to_first_token_seconds_sum"], 0)
-        self.assertEqual(metrics_dict["vllm:time_to_first_token_seconds_bucket"], 3)
+        self.assertEqual(
+            metrics_dict["vllm:time_to_first_token_seconds_bucket"], total_prompts
+        )
         # vllm:time_per_output_token_seconds
         self.assertEqual(metrics_dict["vllm:time_per_output_token_seconds_count"], 45)
         self.assertGreater(metrics_dict["vllm:time_per_output_token_seconds_sum"], 0)
         self.assertEqual(metrics_dict["vllm:time_per_output_token_seconds_bucket"], 45)
         # vllm:e2e_request_latency_seconds
-        self.assertEqual(metrics_dict["vllm:e2e_request_latency_seconds_count"], 3)
+        self.assertEqual(
+            metrics_dict["vllm:e2e_request_latency_seconds_count"], total_prompts
+        )
         self.assertGreater(metrics_dict["vllm:e2e_request_latency_seconds_sum"], 0)
-        self.assertEqual(metrics_dict["vllm:e2e_request_latency_seconds_bucket"], 3)
+        self.assertEqual(
+            metrics_dict["vllm:e2e_request_latency_seconds_bucket"], total_prompts
+        )
         # vllm:request_prompt_tokens
-        self.assertEqual(metrics_dict["vllm:request_prompt_tokens_count"], 3)
+        self.assertEqual(
+            metrics_dict["vllm:request_prompt_tokens_count"], total_prompts
+        )
         self.assertEqual(metrics_dict["vllm:request_prompt_tokens_sum"], 18)
-        self.assertEqual(metrics_dict["vllm:request_prompt_tokens_bucket"], 3)
+        self.assertEqual(
+            metrics_dict["vllm:request_prompt_tokens_bucket"], total_prompts
+        )
         # vllm:request_generation_tokens
-        self.assertEqual(metrics_dict["vllm:request_generation_tokens_count"], 3)
+        self.assertEqual(
+            metrics_dict["vllm:request_generation_tokens_count"], total_prompts
+        )
         self.assertEqual(metrics_dict["vllm:request_generation_tokens_sum"], 48)
-        self.assertEqual(metrics_dict["vllm:request_generation_tokens_bucket"], 3)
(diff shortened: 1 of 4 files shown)


ci/L0_backend_vllm/metrics_test/vllm_metrics_test.py
yinggeh, line 1: I think we need more complex requests to verify the following new metrics. Do you know what they mean? @oandreeva-nv
```
# Histogram of the best_of request parameter.
histogram_best_of_request
# Histogram of the n request parameter.
histogram_n_request
```

ci/L0_backend_vllm/metrics_test/vllm_metrics_test.py
```diff
@@ -118,21 +118,61 @@ def test_vllm_metrics(self):
...
+        )
+        self.assertEqual(metrics_dict["vllm:request_generation_tokens_sum"], 48)
+        self.assertEqual(
+            metrics_dict["vllm:request_generation_tokens_bucket"], total_prompts
+        )
+        # vllm:request_params_best_of
+        self.assertEqual(
+            metrics_dict["vllm:request_params_best_of_count"], total_prompts
+        )
+        self.assertEqual(metrics_dict["vllm:request_params_best_of_sum"], 3)
```
yinggeh, line 168: I think we need more complex requests to verify the following new metrics. Do you know what they mean? @oandreeva-nv

```
# Histogram of the best_of request parameter.
histogram_best_of_request
# Histogram of the n request parameter.
histogram_n_request
```

oandreeva-nv: does vLLM define them anywhere?

src/utils/metrics.py
```diff
@@ -110,16 +138,43 @@ def __init__(self, labels):
                 ],
             )
         )
+        # Request stats
+        #   Latency
+        self.histogram_e2e_time_request = self.histogram_e2e_time_request_family.Metric(
```
oandreeva-nv, line 143: Do we need to think about a way to let the user choose what metrics they want to collect, or we just enforcing all of them?

src/utils/metrics.py
```diff
@@ -172,3 +228,21 @@ def log(self, stats: VllmStats) -> None:
             self.metrics.histogram_time_per_output_token,
             stats.time_per_output_tokens_iter,
         )
+        # Request stats
```
oandreeva-nv, line 231: I would suggest to refactor this as follows:
```python
triton_metrics = [ self.metrics.histogram_time_to_first_token, ... ]
vllm_stats = [ stats.time_to_first_tokens_iter, ... ]

for tuple in zip(triton_metrics, vllm_stats):
    self._log_histogram(tuple)
```
This way it is more extendable in my mind. Triton metrics are all defined in one place, as well as vLLM's stats. 
Potentially, maybe there's a way to extract all Triton metrics and vLLM stats without explicitly composing lists. I would recommend checking

oandreeva-nv: More on `zip`: https://realpython.com/python-zip-function/

ci/L0_backend_vllm/metrics_test/vllm_metrics_test.py
yinggeh: NVM. I found the definition here https://github.com/vllm-project/vllm/blob/v0.5.3.post1/vllm/sampling_params.py#L43-L48

src/utils/metrics.py
yinggeh: I think it's fine as long as no obvious performance drop. Users can use `curl localhost:8002/metrics | grep vllm:xxx` to gather metrics they need. Meanwhile I might start on the optimization after this PR.

src/utils/metrics.py
yinggeh: Addressed

src/utils/metrics.py
rmccorm4: My $0.02, I think it's fine to start with the full set of vllm metrics we collect as binary on/off. We can consider enabling/disabling subsets upon future request or if we find significant issues with any particular metrics later on.

ci/L0_backend_vllm/metrics_test/vllm_metrics_test.py
```diff
@@ -112,27 +112,81 @@ def vllm_infer(
...
+            metrics_dict["vllm:time_to_first_token_seconds_bucket"], total_prompts
+        )
         # vllm:time_per_output_token_seconds
         self.assertEqual(metrics_dict["vllm:time_per_output_token_seconds_count"], 45)
         self.assertGreater(metrics_dict["vllm:time_per_output_token_seconds_sum"], 0)
         self.assertEqual(metrics_dict["vllm:time_per_output_token_seconds_bucket"], 45)
+        # vllm:e2e_request_latency_seconds
+        self.assertEqual(
+            metrics_dict["vllm:e2e_request_latency_seconds_count"], total_prompts
+        )
```
rmccorm4, line 153: nit: I think it would be easier to read/review/maintain this function if we could group the common parts:
```python
histogram_metrics = ["vllm:e2e_request_latency_seconds", "..."]
for metric in metrics:
  # Expect exactly one observation and bucket per prompt
  self.assertEqual(f"{metric}_count", total_prompts)
  self.assertEqual(f"{metric}_bucket", total_prompts)
  
  if metric.endswith("_best_of"):
    self.assertEqual(f"{metric}_sum", best_of*total_prompts)
  elif metric.endswith("_n"):
    self.assertEqual(f"{metric}_sum", n*total_prompts)
  else:
    self.assertGreater(f"{metric}_sum", 0)
```
for as many of the metrics that make sense to follow the same pattern. 

We can have separate/special cases for the metrics that don't fit this pattern.

Feel free to modify or change if any of the above is incorrect, just an example.

rmccorm4 (commented):
LGTM overall, just a nit on cleaning up the testing if you agree with the comment

ci/L0_backend_vllm/metrics_test/vllm_metrics_test.py
yinggeh: Thanks for your input. I personally don't think it's maintainable to add new metric tests since there are a lot of special cases. Counter example
```
# vllm:time_per_output_token_seconds
self.assertEqual(metrics_dict["vllm:time_per_output_token_seconds_count"], 45)
# This line is fine
self.assertGreater(metrics_dict["vllm:time_per_output_token_seconds_sum"], 0)
self.assertEqual(metrics_dict["vllm:time_per_output_token_seconds_bucket"], 45)
```

rmccorm4 (approved):

PR #63, opened 2024-09-06
Update README.md

yinggeh:
List metrics in `vllm:*` instead of the variable name.

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,24 +216,24 @@ curl localhost:8002/metrics
 VLLM stats are reported by the metrics endpoint in fields that are prefixed with
 `vllm:`. Triton currently supports reporting of the following metrics from vLLM.
 ```bash
-# Number of prefill tokens processed.
-counter_prompt_tokens
-# Number of generation tokens processed.
-counter_generation_tokens
+# Counter of prefill tokens processed.
+vllm:prompt_tokens_total
+# Counter of generation tokens processed.
+vllm:generation_tokens_total
 # Histogram of time to first token in seconds.
-histogram_time_to_first_token
+vllm:time_to_first_token_seconds
 # Histogram of time per output token in seconds.
-histogram_time_per_output_token
+vllm:time_per_output_token_seconds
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -216,24 +216,24 @@ curl localhost:8002/metrics
 VLLM stats are reported by the metrics endpoint in fields that are prefixed with
 `vllm:`. Triton currently supports reporting of the following metrics from vLLM.
 ```bash
-# Number of prefill tokens processed.
-counter_prompt_tokens
-# Number of generation tokens processed.
-counter_generation_tokens
+# Counter of prefill tokens processed.
```
rmccorm4, line 219: Do we need both this section and the one right below it? Why not just use the below section instead?

VLLM stats are reported by the metrics endpoint in fields that are prefixed with
`vllm:`. For example, the metrics reported by Triton will like similar to the following:
```bash
# HELP vllm:prompt_tokens_total Number of prefill tokens processed.
# TYPE vllm:prompt_tokens_total counter
vllm:prompt_tokens_total{model="vllm_model",version="1"} 10
# HELP vllm:generation_tokens_total Number of generation tokens processed.
# TYPE vllm:generation_tokens_total counter
vllm:generation_tokens_total{model="vllm_model",version="1"} 16
...
```

yinggeh: I'd rather have another section listing all supported metrics. The sample metrics output is just hard to follow IMO.

yinggeh: And for example, changing `counter_prompt_tokens` to `vllm:prompt_tokens_total` allows reader to easily locate the corresponding output from below section.

rmccorm4: @oandreeva-nv do you mind driving this PR if you have time?

oandreeva-nv: How about we create a dedicated doc for metrics? So that our front facing README is nice and non-cluttered ?

yinggeh:
There will be changes to the vLLM backend metrics in the upcoming releases. Converting this PR to draft. @oandreeva-nv @rmccorm4

PR #65, opened 2024-09-18
vLLM multi gpu tests adjustments

oandreeva-nv:
TLDR: we were writing strings in json, when they should not be strings.


src/model.py
```diff
@@ -187,9 +187,19 @@ def init_engine(self):
...
+            (
+                isinstance(self.vllm_engine_config["enable_lora"], str)
+                and self.vllm_engine_config["enable_lora"].lower() == "true"
+            )
+            or (
+                (
+                    isinstance(self.vllm_engine_config["enable_lora"], bool)
+                    and self.vllm_engine_config["enable_lora"]
+                )
+            )
```
kthui, line 202:
```suggestion
        if "enable_lora" in self.vllm_engine_config.keys() and str(self.vllm_engine_config["enable_lora"]).lower() == "true":
```

I think it can be simplified, because
```
>>> str(True).lower() == "true"
True
>>> str(False).lower() == "true"
False
>>> str("true").lower() == "true"
True
>>> str("false").lower() == "true"
False
>>>
```

kthui (approved):
Thanks for adjusting the boolean variables encoded in strings issue! LGTM overall!

src/model.py
oandreeva-nv: Great suggestion! thanks

kthui (approved):

PR #73, opened 2024-11-19
feat: Auto unload model if vLLM health check failed

kthui:
#### What does the PR do?
Call the vLLM Health Check API upon receiving each request, and auto unload the model via BLS if the health check failed.

#### Checklist

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [x] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Related PRs:
N/A

Start with the documentation, and then move on to the code, and finally the test case.

#### Test plan:
New tests are added to verify the health check, auto model unload, and enabling/disabling of the check.

- CI Pipeline ID: 20829546

#### Caveats:
N/A

#### Background
N/A

N/A


src/model.py
```diff
@@ -131,6 +134,31 @@ def initialize(self, args):
...
+    def _setup_health_check(self):
+        # Check if health check should be enabled
+        self._enable_health_check = (
+            "ENABLE_VLLM_HEALTH_CHECK" in self.model_config["parameters"]
+        ) and (
+            self.model_config["parameters"]["ENABLE_VLLM_HEALTH_CHECK"][
+                "string_value"
+            ].lower()
+            in ["yes", "true"]
+        )
```
rmccorm4, line 146: nit: Can we move these types of checks into a helper function or pb_utils?

ex: 
```python
    def _get_bool_config_param(self, parameter: str) -> bool:
        if parameter in self.model_config["parameters"]
          and (
            self.model_config["parameters"][parameter][
                "string_value"
            ].lower()
            in ["yes", "true"]
        ):
            return True

        return False

...
    self._enable_health_check = self._get_bool_config_param("ENABLE_VLLM_HEALTH_CHECK")
...
    self._report_custom_metrics = self._get_bool_config_param("REPORT_CUSTOM_METRICS")
```

src/model.py
```diff
@@ -131,6 +134,31 @@ def initialize(self, args):
...
+        # Setup health check if enabled
+        if self._enable_health_check:
+            # Only enable health check if there is exactly 1 instance
+            num_instances = 0
+            for group in self.model_config["instance_group"]:
+                num_instances += group["count"]
+            if num_instances != 1:
+                self.logger.log_warn(
+                    f"[vllm] Health check may only be enabled when the model has exactly 1 instance but {num_instances} are found"
+                )
```
rmccorm4, line 156: If the feature is opt-in via config, I don't think we need to restrict it to 1 instance. We can document that the whole model will be unloaded, and they should be aware of this when opting-in to the feature. 

With this check, we'd be limiting people who actually want to do the unload after opting in but have more than 1 instance.

src/model.py
kthui: Sure, refactored here: [Allow health check to be enableable for all instance counts](https://github.com/triton-inference-server/vllm_backend/pull/73/commits/33367f64b7c0cdcb8002194303edae46a106cc56)

src/model.py
kthui: Yes, updated [Allow health check to be enableable for all instance counts](https://github.com/triton-inference-server/vllm_backend/pull/73/commits/33367f64b7c0cdcb8002194303edae46a106cc56)

src/model.py
```diff
@@ -542,6 +570,28 @@ def verify_loras(self, request):
...
+        except Exception as e:
+            self.logger.log_error(f"[vllm] Engine is not healthy: {e}")
+            pb_utils.unload_model(self.model_config["name"])  # non-blocking
+            self._is_healthy = False
+        if not self._is_healthy:
+            for request in requests:
+                request.get_response_sender().send(
+                    pb_utils.InferenceResponse(
+                        error=pb_utils.TritonError(
+                            message="vLLM engine is not healthy",
```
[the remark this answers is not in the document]
kthui: [Update unhealthy unload message](https://github.com/triton-inference-server/vllm_backend/pull/73/commits/6946b8965ab0f5d6e18515fe69a3725b96bfbda1)

src/model.py
```diff
@@ -542,6 +552,30 @@ def verify_loras(self, request):
...
+    def _check_health(self, requests):
+        coro = self.llm_engine.check_health()
+        future = asyncio.run_coroutine_threadsafe(coro, self._loop)
+        try:
+            future.result()
+        except Exception as e:
+            self.logger.log_error(
+                f"[vllm] Engine is not healthy and model will be unloaded: {e}"
+            )
+            pb_utils.unload_model(self.model_config["name"])  # non-blocking
```
rmccorm4, line 564: Technical question out of curiosity - not asking you to implement it here.

Can a python model reload itself? For example, would `pb_utils.load_model(self.model_config["name"])` work after `pb_utils.unload_model(self.model_config["name"])`?

It seems like some python logic within the model still runs after triggering the unload, and I'm curious what the expected behavior is here in general after a model calls unload on itself.

kthui: I think it will work, as long as the BLS load is async, but a hack on the model file timestamp is needed to stop the core from running the update routine which will not reload anything.

kthui: > It seems like some python logic within the model still runs after triggering the unload, and I'm curious what the expected behavior is here in general after a model calls unload on itself.

The unload is async, so it will begin immediately. The model lifecycle will unload / forget the model, but it only [releases](https://github.com/triton-inference-server/core/blob/main/src/model_repository_manager/model_lifecycle.h#L254) the [shared pointer](https://github.com/triton-inference-server/core/blob/main/src/model_repository_manager/model_lifecycle.h#L254) which is also [held by active inference requests](https://github.com/triton-inference-server/core/blob/main/src/infer_request.h#L798), so the model technically will not begin to unload until all active inference requests are completed.

docs/health_check.md
```diff
@@ -0,0 +1,58 @@
...
+-->
+
+# vLLM Health Check (BETA)
+
+> [!NOTE]
+> The vLLM Health Check support is currently in BETA. Its features and
+> functionality are subject to change as we collect feedback. We are excited to
+> hear any thoughts you have!
+
+The vLLM backend supports checking for
```
oandreeva-nv, line 36:
```suggestion
Starting with r24.12 vLLM backend supports checking for
```

src/model.py
```diff
@@ -675,6 +685,30 @@ def verify_loras(self, request):
...
+            self._is_healthy = False
+        if not self._is_healthy:
+            for request in requests:
+                request.get_response_sender().send(
+                    pb_utils.InferenceResponse(
+                        error=pb_utils.TritonError(
+                            message="vLLM engine is not healthy and model will be unloaded",
```
oandreeva-nv, line 704: this goes to user, right? If so, I think user might be confused with loaded and unloaded terms

kthui: Why the user is confused?

oandreeva-nv: First question if I'm unaware of Triton, what does that mean? Maybe something like: "... model is currently unavailable " ? What do you think?

docs/health_check.md
```diff
@@ -0,0 +1,58 @@
...
+[vLLM Engine Health](https://github.com/vllm-project/vllm/blob/v0.6.3.post1/vllm/engine/async_llm_engine.py#L1177-L1185)
+upon receiving each inference request. If the health check fails, the entire
+model will be unloaded, so its state becomes NOT Ready at the server, which can
+be queried by the
+[Repository Index](https://github.com/triton-inference-server/server/blob/main/docs/protocol/extension_model_repository.md#index)
+or
+[Model Ready](https://github.com/triton-inference-server/client/blob/main/src/c%2B%2B/library/http_client.h#L178-L192)
+APIs.
+
+The Health Check is disabled by default. To enable it, set the following
```
oandreeva-nv, line 46: Do we expect any perf impacts, or they are negligible ?

docs/health_check.md
```diff
@@ -0,0 +1,58 @@
...
+
+> [!NOTE]
+> The vLLM Health Check support is currently in BETA. Its features and
+> functionality are subject to change as we collect feedback. We are excited to
+> hear any thoughts you have!
+
+The vLLM backend supports checking for
+[vLLM Engine Health](https://github.com/vllm-project/vllm/blob/v0.6.3.post1/vllm/engine/async_llm_engine.py#L1177-L1185)
+upon receiving each inference request. If the health check fails, the entire
+model will be unloaded, so its state becomes NOT Ready at the server, which can
```
GuanLuo, line 39: "unload" is a bit related to implementation detail, users only need to know the model will be not ready
```suggestion
model state will become NOT Ready at the server, which can
```

src/model.py
```diff
@@ -229,6 +233,12 @@ def init_engine(self):
                 else:
                     raise e
 
+    def _get_bool_config_param(self, param_name: str) -> bool:
+        return (param_name in self.model_config["parameters"]) and (
+            self.model_config["parameters"][param_name]["string_value"].lower()
+            in ["yes", "true"]
```
GuanLuo, line 239: nit: make sure we don't actually use "yes" in our example.. We can parse it but we should use "true/false" ("enable/disable"?) for consistency

src/model.py
```diff
@@ -675,6 +685,30 @@ def verify_loras(self, request):
             verified_request = request
         return verified_request
 
+    def _check_health(self, requests):
+        coro = self.llm_engine.check_health()
+        future = asyncio.run_coroutine_threadsafe(coro, self._loop)
+        try:
+            future.result()
```
GuanLuo, line 692: How is this different from `await self.llm_engine.check_health()`?

docs/health_check.md
kthui: I would not mention the supported version at the beginning, because the user is simply checking out what this feature is, and the version is only relevant when they want to try it out.

docs/health_check.md
```diff
@@ -0,0 +1,58 @@
...
+parameters: {
+  key: "ENABLE_VLLM_HEALTH_CHECK"
+  value: { string_value: "true" }
+}
+```
+and select
+[Model Control Mode EXPLICIT](https://github.com/triton-inference-server/server/blob/main/docs/user_guide/model_management.md#model-control-mode-explicit)
+when the server is started.
+
+Supported since r24.12.
```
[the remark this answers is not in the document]
kthui: Second though. I think we do not need to mention the version, because the feature is supported iff this documentation present.

src/model.py
kthui: Ok. The user does not need to know if the model will be unloaded, which is an implementation detail. How about we change it to "Model is unavailable due to unhealthy vLLM engine"?

docs/health_check.md
kthui: We should not expect any impact if the feature is disabled. I will benchmark the numbers when it is enabled.

docs/health_check.md
kthui: Yes, it is implementation details. Updated.

docs/health_check.md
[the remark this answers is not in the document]
oandreeva-nv: Sure, let's delete this atogether

src/model.py
kthui: I am more than happy to allow only "true", but "yes" is a legacy from the [metrics implementation](https://github.com/triton-inference-server/vllm_backend/blob/r24.11/src/model.py#L167-L168).

I think there are two ways to resolve it:
1. Have the metrics parse the bool value differently than the health check, but this will introduce inconsistency within the vLLM backend.
2. Update the metrics bool parsing to accept "true" only, but this can be a breaking change for existing users.

docs/health_check.md
oandreeva-nv: Sure, I understand this, I was wondering if it make sense to notify users that perf hit is expected, when this feature is enabled

src/model.py
nnshah1: as its a "bool config param" suggest "true/false" only

src/model.py
kthui: [Update docs and messages](https://github.com/triton-inference-server/vllm_backend/pull/73/commits/b6bd6496627f7eb37ab94899a5eda98d3ffef48d)

src/model.py
nnshah1: I guess I would be in favor of breaking compat - to my knowledge we don't use 'yes' anywhere else ....

docs/health_check.md
kthui: > I was wondering if it make sense to notify users that perf hit is expected, when this feature is enabled

Yes, I will mention it in the docs.

> can we benchmark both?

Sure. I will run the benchmarking once we are good with the change.

src/model.py
kthui: sure, updated bool config param to accept true/false only
https://github.com/triton-inference-server/vllm_backend/pull/73/commits/a26a083755da584ce3b117e09c907d4b0f90c71c

src/model.py
oandreeva-nv: > so the model technically will not begin to unload until all active inference requests are completed.

How are they going to be completed if the engine is unhealthy?

docs/health_check.md
kthui: Ran the benchmarking. Let us sync-up offline.

src/model.py
kthui: > How are they going to be completed if the engine is unhealthy?

This is an interesting point. To complete an inference request, all we need is to send the complete final flag. In other words, we don't need to pass the requests into the engine, but respond with an error and complete final flag as soon as they arrive at the model (i.e. those queued in the scheduler). We will also need a way to track the requests processing by the engine and respond to them as well.

Back to the original question, "can a python model reload itself?". The unload of the unhealthy model does not need to be completed before reloading and serving the reloaded the model. In other words, the unhealthy model and the reloaded model can coexist.

GuanLuo (approved):

PR #75, opened 2024-12-03
feat: Add log probabilities and number of input tokens to additional outputs

kthui:
#### Checklist
- [ ] PR title reflects the change and is of format `<commit_type>: <Title>`
- [ ] Changes are described in the pull request.
- [ ] Related issues are referenced.
- [ ] Added [test plan](#test-plan) and verified test passes.
- [ ] Verified that the PR passes existing CI.
- [ ] Verified copyright is correct on all changed files.

#### Commit Type:
- [ ] build
- [ ] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Test plan:

- CI Pipeline ID:

- closes GitHub issue: #xxx


kthui:
For example: 

text_output: " the"
logprobs: [{'5': {'logprob': -1.0246853828430176, 'rank': 1, 'decoded_token': ' the'}, '10': {'logprob': -2.2121853828430176, 'rank': 2, 'decoded_token': ' a'}}]

text_output: " term"
logprobs: [{'1385': {'logprob': -3.2646501064300537, 'rank': 1, 'decoded_token': ' term'}, '78': {'logprob': -3.7529313564300537, 'rank': 2, 'decoded_token': ' first'}}]

ci/L0_additional_outputs_vllm/additional_outputs_test.py
```diff
@@ -76,6 +84,16 @@ def _get_inputs(
                 np.array([return_cumulative_logprob], dtype=bool)
             )
 
+        if return_logprobs is not None:
+            inputs.append(grpcclient.InferInput("return_logprobs", [1], "BOOL"))
```
rmccorm4, line 88: Let's be consistent with TRTLLM and use `return_log_probs`: https://github.com/triton-inference-server/tensorrtllm_backend/blob/91c07d3acd865238176f0fcc3375155e3085c5c8/all_models/inflight_batcher_llm/ensemble/config.pbtxt#L442

src/model.py
```diff
@@ -104,6 +104,18 @@ def _auto_complete_inputs_and_outputs(auto_complete_model_config):
                 "dims": [1],
                 "optional": True,
             },
+            {
+                "name": "return_logprobs",
```
rmccorm4, line 108: Let's be consistent with TRTLLM and use return_log_probs for the input name: https://github.com/triton-inference-server/tensorrtllm_backend/blob/91c07d3acd865238176f0fcc3375155e3085c5c8/all_models/inflight_batcher_llm/ensemble/config.pbtxt#L442

src/model.py
```diff
@@ -496,21 +513,58 @@ def _create_response(
...
+                request_output_state["prev_lens_logprobs"][i] = len(output.logprobs)
+                logprobs_py = []
+                for logprob_d_vllm in output.logprobs[prev_len:]:
+                    logprob_d_py = {}
+                    for token_id, logprob_vllm in logprob_d_vllm.items():
+                        logprob_d_py[token_id] = {
+                            "logprob": logprob_vllm.logprob,
+                            "rank": logprob_vllm.rank,
+                            "decoded_token": logprob_vllm.decoded_token,
+                        }
```
rmccorm4, line 538: What's the reference for this implementation? How did you decide to return rank, decoded_token, etc. and how many top N logprobs to return?

src/model.py
kthui: Should we also unify the output names?
https://github.com/triton-inference-server/tensorrtllm_backend/blob/91c07d3acd865238176f0fcc3375155e3085c5c8/all_models/inflight_batcher_llm/ensemble/config.pbtxt#L505-L512

The content of vLLM logprobs is different from TRT-LLM. It might not be a bad idea to keep them different to distinguish between their difference.

src/model.py
kthui: > What's the reference for this implementation?
https://github.com/vllm-project/vllm/blob/main/vllm/sequence.py#L53-L54

> How did you decide to return rank, decoded_token, etc
They are vLLM specific from https://github.com/vllm-project/vllm/blob/main/vllm/sequence.py#L33-L47

> how many top N logprobs to return?
This is controlled by the user by setting the input parameter `logprobs`. vLLM returns up to logprobs + 1. https://github.com/vllm-project/vllm/blob/main/vllm/sampling_params.py#L143-L149

docs/additional_outputs.md
```diff
@@ -57,7 +55,26 @@ for more details.
 To enable, set `return_cumulative_logprob` input tensor to `True`. The floating
 point value will be sent on the `cumulative_logprob` output tensor.
 
-Supported since r24.12.
+### Log Probabilities
+
+The log probabilities of the top probability words at each position if the
```
rmccorm4, line 60:
```suggestion
The log probabilities of the top probability tokens at each position if the
```

docs/additional_outputs.md
```diff
@@ -57,7 +55,26 @@ for more details.
...
+[logprobs](https://github.com/vllm-project/vllm/blob/v0.6.3.post1/vllm/sampling_params.py#L146-L152)
+are requested. Only the log probabilities of the new tokens generated since the
+last response are returned on each new response. See
+[here](https://github.com/vllm-project/vllm/blob/v0.6.3.post1/vllm/outputs.py#L24-L25)
+for more details on the log probabilities.
+
+To enable, set `return_logprobs` input tensor to `True`. The log probabilities
+will be sent on the `logprobs` output tensor as JSON.
```
rmccorm4, line 68:
```suggestion
will be sent on the `logprobs` output tensor as a serialized JSON string.
```

rmccorm4 (commented):
Minor doc comments - LGTM otherwise

rmccorm4 (approved):
@oandreeva-nv for second opinion

src/model.py
```diff
@@ -496,21 +513,58 @@ def _create_response(
                 )
             )
 
+        # logprobs
```
oandreeva-nv, line 516: nit: Let's try to provide more descriptive comments, i.e. "Handling logprobs outputs (+ link on vllm doc)"

src/model.py
```diff
@@ -496,21 +513,58 @@ def _create_response(
...
+                            "decoded_token": logprob_vllm.decoded_token,
+                        }
+                    logprobs_py.append(logprob_d_py)
+                logprobs.append(json.dumps(logprobs_py).encode("utf-8"))
+            output_tensors.append(
+                pb_utils.Tensor("logprobs", np.asarray(logprobs, dtype=np.object_))
+            )
+
+        # num_input_tokens
+        if additional_outputs["return_num_input_tokens"]:
```
oandreeva-nv, line 546: is "return_num_input_tokens" guaranteed to be in `additional_outputs`?

kthui: guaranteed here:
https://github.com/triton-inference-server/vllm_backend/pull/75/files#diff-6159e4e86b1b15ffd250a78dbdf29a41b004b4963154e49f8b96609b8ea977ceR398-R399

src/model.py
kthui: It is a good idea to add the link, so it is easier to lookup the vLLM implementation in the future.

I think it would be redundant to add "handling" and "outputs" to the comment, because the method name "_create_response" already imply handling outputs, and the only missing information here is which output is this block/section is handling, which I set the comment to match the output tensor name exactly to be very clear that this block is handling logprobs outputs. What do you think?

oandreeva-nv: Sure, I'm ok with just adding link.

PR #78, opened 2024-12-20
Setting shutdown asyncio event in a thread-safe manner

oandreeva-nv:
It seems like we need to be careful, when using asyncio with multi-threading. The non-graceful shutdown was happening due to "RuntimeError: Non-thread-safe operation invoked on an event loop other than the current one\n\nA" (I started `asyncio.run` with `debug=True`). Thus I'm replacing asyncio with threading.

Tests: https://gitlab-master.nvidia.com/dl/dgx/tritonserver/-/jobs/129814805
tests ran for branch `oandreeva_metrics_refactor`, but the change is the same. I suggest to merge this commit to the current perf PR and merge perf PR to main, i.e. not wait for metrics follow up.

### DIFF
--- a/ci/L0_backend_vllm/metrics_test/test.sh
+++ b/ci/L0_backend_vllm/metrics_test/test.sh
@@ -75,11 +75,11 @@ run_test() {
         fi
     fi
 
+    set -e
+
     # TODO: Non-graceful shutdown when metrics are enabled.
     kill $SERVER_PID
     wait $SERVER_PID
-
-    set -e
 }
 
 RET=0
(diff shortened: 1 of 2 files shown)


kthui:
Can you give this a quick try?
```python
def finalize(self):
    ...
    self._event_loop.call_soon_threadsafe(self._llm_engine_shutdown_event.set)
    ...
```

It let the asyncio event to be set in the same loop that the event is blocked on, so we don't need a "busy" timeout loop for checking the shutdown event.

kthui (approved):

PR #93, opened 2025-07-28
Mchornyi/tprd 1641/lock transformers version

mc-nv:


yinggeh (approved):
Probably need a plan to keep transformers up-to-date once vLLM has an update.

PR #97, opened 2025-09-26
Update L0_multi_gpu_vllm script in order to be able access resources …

mc-nv:
…within container. Those resources no longer shipped with 'pytorch-triton' wheel package.

### DIFF
--- a/ci/L0_multi_gpu_vllm/multi_lora/test.sh
+++ b/ci/L0_multi_gpu_vllm/multi_lora/test.sh
@@ -41,6 +41,9 @@ EXPECTED_NUM_TESTS=2
 GENERATE_ENDPOINT="localhost:8000/v2/models/vllm_llama_multi_lora/generate"
 CHECK_FOR_ERROR=true
 
+export C_INCLUDE_PATH=/usr/local/cuda/include:$C_INCLUDE_PATH
+export TRITON_PTXAS_PATH=/usr/local/cuda/bin/ptxas
+
 make_api_call() {
     local endpoint="$1"
     local data="$2"
--- a/ci/L0_multi_gpu_vllm/vllm_backend/test.sh
+++ b/ci/L0_multi_gpu_vllm/vllm_backend/test.sh
@@ -36,6 +36,9 @@ CLIENT_PY="./vllm_multi_gpu_test.py"
 SAMPLE_MODELS_REPO="../../../samples/model_repository"
 EXPECTED_NUM_TESTS=1
 
+export C_INCLUDE_PATH=/usr/local/cuda/include:$C_INCLUDE_PATH
+export TRITON_PTXAS_PATH=/usr/local/cuda/bin/ptxas
+
 ### Helpers
 function validate_file_contains() {
     local KEY="${1}"


ci/L0_multi_gpu_vllm/vllm_backend/test.sh
```diff
@@ -36,6 +36,9 @@ CLIENT_PY="./vllm_multi_gpu_test.py"
 SAMPLE_MODELS_REPO="../../../samples/model_repository"
 EXPECTED_NUM_TESTS=1
 
+export C_INCLUDE_PATH=/usr/local/cuda/include:$C_INCLUDE_PATH
```
yinggeh, line 39: Does pytorch_triton drops these two files perpetually? If not, please add a comment associates with a ticket and remove them once they fix the bug.

ci/L0_multi_gpu_vllm/multi_lora/test.sh
```diff
@@ -41,6 +41,9 @@ EXPECTED_NUM_TESTS=2
 GENERATE_ENDPOINT="localhost:8000/v2/models/vllm_llama_multi_lora/generate"
 CHECK_FOR_ERROR=true
 
+export C_INCLUDE_PATH=/usr/local/cuda/include:$C_INCLUDE_PATH
+export TRITON_PTXAS_PATH=/usr/local/cuda/bin/ptxas
```
dmitry-tokarev-nv, line 45: @mc-nv can you add comments. So we know later why these were added?

dmitry-tokarev-nv: commit msgs get buries when files are moved/refactored.

ci/L0_multi_gpu_vllm/multi_lora/test.sh
```diff
@@ -41,6 +41,9 @@ EXPECTED_NUM_TESTS=2
 GENERATE_ENDPOINT="localhost:8000/v2/models/vllm_llama_multi_lora/generate"
 CHECK_FOR_ERROR=true
```
dmitry-tokarev-nv, line 43:
```suggestion

# These resources no longer shipped with 'pytorch-triton' wheel package.
```

ci/L0_multi_gpu_vllm/vllm_backend/test.sh
```diff
@@ -36,6 +36,9 @@ CLIENT_PY="./vllm_multi_gpu_test.py"
 SAMPLE_MODELS_REPO="../../../samples/model_repository"
 EXPECTED_NUM_TESTS=1
```
dmitry-tokarev-nv, line 38:
```suggestion

# These resources no longer shipped with 'pytorch-triton' wheel package.
```

ci/L0_multi_gpu_vllm/multi_lora/test.sh
dmitry-tokarev-nv: added suggestions

dmitry-tokarev-nv (approved):

ci/L0_multi_gpu_vllm/multi_lora/test.sh
mc-nv: I would like to clarify with you which course of actions I should take on change. Because, you lift/dismiss the request for change on PR but bring the suggestions. Nest to it you provide approval. 
I'm confused, on further actions as I can't apply the suggestions because they are in irrelevant place.
 
[`C_INCLUDE_PATH`](https://gcc.gnu.org/onlinedocs/cpp/Environment-Variables.html) - is as system level designed variable used by `gcc` compiler to look up for. 

[`TRITON_PTXAS_PATH`](https://github.com/triton-lang/triton/blob/main/python/triton/knobs.py#L199) - is a variable that has normalized name and has corresponded reference to the package that is been used and installed within container. 

With above said I don't know if you would like me to apply your suggestions before merge?

PR #128, opened 2026-04-22
test: Fix various vLLM tests

yinggeh:
Fix two CI failures in vLLM backend CI tests.                                                     
  
  1. **`L0_backend_vllm` streaming test**: Refactors `_test_vllm_model` to                          
     verify the *joined* complete output string (`expected_joined_output`) rather
     than comparing against a hardcoded list of individual streamed chunks. Adds                    
     an assertion that the stream actually delivers more than one chunk (ensuring
     streaming is exercised). Removes the hardcoded error-message string check for                  
     the `exclude_input_in_output=False` case — the test already asserts                            
     `expect_error=True`, which is sufficient.                                                      
                                                                                                    
  2. **`L0_multi_gpu_vllm` test script**: Pins `ray==2.55.0` to fix import                          
     failures caused by an incompatible `ray` version being pulled in during CI.

### DIFF
--- a/ci/L0_backend_vllm/enabled_stream/enabled_stream_test.py
+++ b/ci/L0_backend_vllm/enabled_stream/enabled_stream_test.py
@@ -1,4 +1,4 @@
-# Copyright 2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright 2023-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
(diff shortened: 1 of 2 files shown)


mudit-eng:
@yinggeh - Can you share a summary of the changes?

yinggeh:
> @yinggeh - Can you share a summary of the changes?

done

mc-nv (approved):

whoisj (approved):

